PR #2, opened 2015-10-07
Fix to "Quick example: Counter", was not outputting any data. 

vildehav:
Now showing current count (not total number of clicks which may have been the intent?).


acdlite:
D'oh, thanks!

PR #3, opened 2015-10-07
Fix to "Quick example: Counter", not working

vildehav:
Removed property "counter" from mapProps (is supplied by ...rest).


acdlite:
D'oh, thanks :D

PR #13, opened 2015-10-10
defaultProps in React overrides props defined with undefined value

istarkov:
Look at failing test for more details.


acdlite:
Can you split this up into two PRs, please? I'm fine with the `setState()` callback, but it occurs to me that `defaultProps()` should just use React's static `defaultProps()` property so that we don't have divergent behavior — I actually though this is what it did, but it must have escaped my notice when I was writing that module.

istarkov:
Done in #14 and #15

PR #14, opened 2015-10-10
defaultProps overrides undefined props from owner

istarkov:
Slit this https://github.com/acdlite/recompose/pull/13
Also move props 2 defaultProps as React do


acdlite:
Thanks!

PR #19, opened 2015-10-14
Allow withState accept initialState as function of props

istarkov:


istarkov:
Also I think the same must be done for defaultProps, if you agree with this
```
@defaultProps(props => ({
...blablabla
}))
```

acdlite:
Thanks!

PR #22, opened 2015-10-20
createSpy() and use stateless function components where possible

acdlite:
Adds a test helper `createSpy()` that provides an alternative to React TestUtils's `scry*` and `find*` methods for tests. A spy is a higher-order component which tracks the props for each render. The spy also has a ref to the wrapped component.

It has the advantage of working with both normal, class components and stateless function components, which React TestUtils currently does not support. This will allow Recompose to implement many of its HoCs as functions rather than class components.

This is used internally to test Recompose's helpers.

Need to add docs before merging.


acdlite:
Here's an example of how to use component spies:

```js
const spy = createSpy();
const Spy = compose(
  withState('n', 'updateN', props => props.initialN),
  spy
)('div');

renderIntoDocument(
  <div>
    <Spy initialN={1} />
    <Spy initialN={3} />
    <Spy initialN={5} />
  </div>
);

spy.getProps(0, 0).updateN(n => n * 7);

expect(spy.getProps().n).to.equal(35); // Test default indices
expect(spy.getProps(0, 0).n).to.equal(35);
expect(spy.getProps(0, 1).n).to.equal(5);
expect(spy.getProps(1, 0).n).to.equal(3);
expect(spy.getProps(2, 0).n).to.equal(1);
```

If the spied component is not a stateless function component, you can also use `getComponent()` to return a ref.

acdlite:
Not sure if this should be released as part of Recompose. Skeptical of its utility for most projects. You'd need to combine it with something like [inject-loader](https://github.com/plasticine/inject-loader). Maybe I'll release it separately as recompose-spy.

acdlite:
Yeah I'm going to keep this private for now.

Issue #18, opened 2015-10-12
Internal state without setState


istarkov:
There are situations at almost all controls I've wrote when I need to hold some state but this state isn't React or Flux state. Because I don't want additional rerender.

For example:

```javascript
  _onMouseEnter(poly) {
    clearTimeout(this.mouseLeaveTimeoutHandle_);
    onHoveredChange(poly);
  }

  _onMouseLeave( /* poly */ ) {
    this.mouseLeaveTimeoutHandle_ = setTimeout(
      () => onHoveredChange(null),
      1000 / 60
    );
  }
```

Here I use  `this.mouseLeaveTimeoutHandle_` to prevent rerender if Mouse enter new object.
Sometimes such `state` used as cache, sometimes to support imperative DOM calls.

May be recompose needs something like `withState` but with `getter` functions support and without `setState` calls?

(For such state Getter Setter functions (or one function) are passed through props, not a value)

istarkov:
I can share such state using 
```javascript
@mapProps(p => {
   let internalState = 0;
   return {
      ...p,
      incInternalState() {
        ++internalState; 
      },
      getInternalState() {
        return internalState; 
      },
   };
})
```
May be this enough.

istarkov:
Look's like enough :-)

Issue #8, opened 2015-10-08
mapPropsOnUpdate add childProps to propsMapper call


istarkov:
Add childProps to propsMapper call here
https://github.com/acdlite/recompose/blob/master/src/mapPropsOnUpdate.js#L19

Or allow additional check

Sometimes i need to do additional checks to decide do i need to recalculate props or can return previous value.

For example

```javascript
@mapPropsOnUpdate(
  ['a', 'b'],
  (props) => {
    if (a<b) { // it can be computationally intensive check 
        return {...props, result: I_WANT_TO_RETURN_PREV_RESULT} 
    }
    return {...props, result: calcNewResult(props)} 
  }
```

with childProps it will be easy

```javascript
@mapPropsOnUpdate(
  ['a', 'b'],
  (props, childProps) => {
    if (a<b) { // it can be computationally intensive check 
        return {...props, result: childProps.result} 
    }
    return {...props, result: calcNewResult(props)} 
  }
```

acdlite:
`mapPropsOnUpdate()` is just a shortcut for `compose(shouldUpdate(...), mapProps(...))`. If you want more control over when to update, use that.

istarkov:
It's not `compose` of `shouldUpdate` and `mapProps`
component may update and not recalculate some internal values.

Your code for `mapPropsOnUpdate` does not contain any `shouldComponentUpdate` calls.

And there is no method like 
`mapPropsOnUpdate(props => myCompareFn, props => newProps) `

acdlite:
Ah you're right, clearly I don't my own library well enough :)

Let me think about this one for a bit.

acdlite:
I know the implementation is slightly different, but is there any reason why `compose(shouldUpdate(...), mapProps(...))` wouldn't achieve the desired affect?

istarkov:
I need to redraw but not to call again some intensive calculate function. 
And in redraw i need result of that function.
And if some props changed, i need to run calculate function again after some additional checks.

Animation - in my current task. 
I have map with svg tiles over it. 
Instead of full svg path points position recalculation i can use css transforms until some props changes.

For example i scale on zoom oversimplified map tiles until tiles for new zoom level arrived.
Tiles preparation for swg drawing is really intensive task.
Also I can't call recalculate until google map stop animation (otherwise it breaks animation)

So i check `shallow equality` on some props inside `componentWillReceiveProps`,
and if it `false` I make additional checks -  could i reuse previous calculation partially or need to make full recalc again.

[...]

PR #27, opened 2015-10-23
Add reduceProps

istarkov:
It looks like it's very useful decorator for my work, 
It also solve this issue https://github.com/acdlite/recompose/issues/8
And this https://github.com/acdlite/recompose/issues/18

### DIFF
--- a/src/reduceProps.js
+++ b/src/reduceProps.js
@@ -1,0 +1,18 @@
+import curry from 'lodash/function/curry';
+import wrapDisplayName from './wrapDisplayName';
+import createElement from './createElement';
+
+const reduceProps = (reduce, initialState, BaseComponent) => {
+  let state = initialState;
+
+  const ReduceProps = props => {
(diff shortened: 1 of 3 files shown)


acdlite:
Thanks for the suggestion. Let me think about it for a bit.

src/reduceProps.js
```diff
@@ -0,0 +1,18 @@
+import curry from 'lodash/function/curry';
+import wrapDisplayName from './wrapDisplayName';
+import createElement from './createElement';
+
+const reduceProps = (reduce, initialState, BaseComponent) => {
+  let state = initialState;
```
acdlite: This value is shared between all instances of ReduceProps. You need it to be local to each component. The only way to do this is with a class component.

istarkov: Ooops, what a stupid bug I made :-(

istarkov:
Look's like I need to rethink this.

PR #28, opened 2015-10-23
Fail test for defaultProps

istarkov:
defaultProps doesn't work in combination with other recompose hoc (only if defined first in chain)

### DIFF
--- a/src/__tests__/defaultProps-test.js
+++ b/src/__tests__/defaultProps-test.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { expect } from 'chai';
-import { defaultProps, compose } from 'recompose';
+import { defaultProps, pure, compose } from 'recompose';
 import createSpy from './createSpy';
 
 import { renderIntoDocument } from 'react-addons-test-utils';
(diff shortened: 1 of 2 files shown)


acdlite:
Ah a better first for this would be to update `isReferentiallyTransparentFunctionComponent()` to check for existence of `defaultProps`.

```js
const isReferentiallyTransparentFunctionComponent = Component => (
  Component &&
  typeof Component !== 'string' &&
  !(Component.prototype && Component.prototype.render) &&
  !Component.defaultProps &&
  !Component.contextTypes
);
```

Want to update the PR to fix that?

acdlite:
Actually I'll handle it, will update this branch and merge. Thanks!

istarkov:
Thank you, 
so nice source code is like a programming book for me.

acdlite:
Ha thanks :)

In case you're wondering, the reason that works is because we have a custom version of `createElement()` that we use internally. If the component is a function and is "referentially transparent" (not sure if that's the best term but what I mean by that is returning `<A/>` and `A()` have the same behavior) then we call the function to avoid an intermediate React element. I added a [section](https://github.com/acdlite/recompose#should-i-use-this-performance-and-other-concerns) to the README that discusses this.

PR #32, opened 2015-10-26
Rename mapPropsOnUpdate -> mapPropsOnChange

acdlite:
The word "update" wrongly implies that `shouldComponentUpdate()` is somehow involved.


istarkov:
Yes!

PR #33, opened 2015-10-29
OnlyUpdateForProps

grabbou:
WIP

### DIFF
--- a/src/onlyUpdateForProps.js
+++ b/src/onlyUpdateForProps.js
@@ -1,0 +1,8 @@
+import onlyUpdateForKeys from 'onlyUpdateForKeys';
+
+const onlyUpdateForProps = (BaseComponent) => {
+  const propKeys = Object.keys(BaseComponent.propTypes || {});
+  return onlyUpdateForKeys(propKeys, BaseComponent);
+};
+
+export default onlyUpdateForProps;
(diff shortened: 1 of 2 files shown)


src/onlyUpdateForProps.js
```diff
@@ -0,0 +1,8 @@
+import onlyUpdateForKeys from 'onlyUpdateForKeys';
+
+const onlyUpdateForProps = (BaseComponent) => {
+  const propKeys = Object.keys(BaseComponent.propTypes || {});
```
acdlite: Hmm, what should the behavior be if the component doesn't have any `propTypes`?

Perhaps we should warn? Otherwise the component will never update. I can't think of any situation where that would be intended behavior.

grabbou: That's right. Maybe we should just return it untouched, e.g `BaseComponent` + optionally print a `warning` since in this case it's clearly an implementation issue.

acdlite: Yeah, I'd say return it untouched and print a warning.

acdlite:
I'm a bit unsure about the name. Seems easy to confuse with `onlyUpdateForKeys()`. What do you think of `onlyUpdateForPropTypes()` instead?

istarkov:
Is it a good idea for HOCs which can be places only as last compose member or must be defined strongly before `setPropTypes` HOC?

acdlite:
@istarkov I think once we add a helpful warning it's okay

Issue #21, opened 2015-10-19
Conditional or Switch HOC?


arturmuller:
Love the work you're doing on this.

Would it makes sense within the context of this library to provide helpers for rendering different components based on props?

For instance `conditional()` could work like `branch()` but instead of applying left/right HOC to the base component, it would render base component if test returns true and null if false.

Or `switch()`, which would get an array of objects: `{ test: (props: Object) => boolean, component: Class<ReactComponent> }` and renders the first one that passes it's test. Base component could act as the `default` in switch statements.

I use something similar to the above on my current project and would be happy to take a crack at a PR for recompose if this is something that makes sense to you.

acdlite:
Thanks for your suggestions! I'm hesitant to add a `conditional()` function because it's not clear from the name how it's different from `branch()`. However, I've been thinking about adding a `renderComponent()` helper, which is really just the identity function with a more meaningful name. With this helper, you could achieve the same effect as `conditional()` like so:

```js
const Component = branch(
  test,
  renderComponent(Foo), // Foo is rendered if true
  renderComponent(Bar) // Bar is rendered if false
)(BaseComponent)
```

This involves a bit more typing, but it avoids the confusion of whether `branch()` accepts an HoC or a component. I'm open to also adding your proposed `conditional()` helper if you can think of a better name for it.

Your `switch()` idea is also neat, but again, I would want it to operate on HoCs, not components.

acdlite:
I've added `renderComponent()` to version 0.5.0. There's an example of how to use it here: https://github.com/acdlite/recompose/blob/master/docs/API.md#rendercomponent

bholloway:
@acdlite I think your resolution was good but there is definitely a place for a `switch()` that operates on HOC. Is there another issue that attends to this?

I would like to handle 4 cases for a prop:

`missing | instanceof Error | empty List | default (non-empty list)`

Currently I need to compose 3 x `branch` HOCs to do this.

smeijer:
Having the same "issue" as @bholloway. I need to render an `Avatar` based on either: 

`userId`: fetch user from api
`user`: should contain all info
`picture`: render simple image
`anonymous`: show guest icon

```jsx
<Avatar userId="u1" />
<Avatar user={{ picture: '...' }} />
<Avatar picture="http://....." />
<Avatar />
```

wuct:
[...]

PR #43, opened 2015-11-06
correct renderComponent() example

mrblueblue:
Since the second argument of `branch` should be a HoC, I've wrapped `BaseComponent` inside a HoC using `renderComponent`.

Also, if `Post` is to be a component, then the curried `branch` function needs to be invoked with a component. In the example, I've used `renderNothing()` as this component.

However, using `branch` like this kinda of feels hacky--especially with using `renderNothing()` as its BaseComponent--but it does produce the desired effect of the `conditional()` mentioned in issue #21.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -214,9 +214,9 @@ This is useful in combination with another helper that expects a higher-order co
 // has all the props it needs
 const spinnerWhileLoading = (hasLoaded, BaseComponent) => branch(
   hasLoaded,
-  BaseComponent,
+  renderComponent(BaseComponent),
   renderComponent(Spinner) // <Spinner> is a React component
-);
+)(renderNothing());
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -214,9 +214,9 @@ This is useful in combination with another helper that expects a higher-order co
 // has all the props it needs
 const spinnerWhileLoading = (hasLoaded, BaseComponent) => branch(
   hasLoaded,
-  BaseComponent,
+  renderComponent(BaseComponent),
   renderComponent(Spinner) // <Spinner> is a React component
-);
+)(renderNothing());
```
acdlite: You actually don't need this line. `spinnerWhileLoading` is a higher-order component, not a component.

PR #52, opened 2015-11-11
Suppress React's warning when setState() is called after unmount

istarkov:
There are a lot of situation we call setState inside some asynchronous methods like promises, callbacks, in direct window events processing.

And component can be unmounted before asynchronous method has completed.

And calling withState method after component has unmounted causes error message in console like 
```
Error: Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the withState(spy(div)) component
```

I've added a test which emulates such behaviour, and fails on any console.error message.

This lines in test looks like a dirty hack, but I don't know a better solution
```javascript
  beforeEach(() => {
    console.error = (message) => {
      throw new Error(message)
    }
  })

  afterEach(() => {
    console.error = savedConsoleError
  })
```

### DIFF
--- a/src/packages/recompose/withState.js
+++ b/src/packages/recompose/withState.js
@@ -16,11 +16,18 @@ export const withState = (
         : initialState
     }
 
+    hasUnmounted = false
+
     updateStateValue = (updateFn, callback) => (
-      this.setState(({ stateValue }) => ({
-        stateValue: isFunction(updateFn) ? updateFn(stateValue) : updateFn
-      }), callback)
+      !this.hasUnmounted &&
+        this.setState(({ stateValue }) => ({
+          stateValue: isFunction(updateFn) ? updateFn(stateValue) : updateFn
+        }), callback)
     )
+
+    componentWillUnmount() {
+      this.hasUnmounted = true
+    }
 
     render() {
       return createElement(BaseComponent, {
--- a/src/packages/recompose/__tests__/withState-test.js
+++ b/src/packages/recompose/__tests__/withState-test.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import { expect } from 'chai'
 import omit from 'lodash/object/omit'
+import pick from 'lodash/object/pick'
 import { withState, compose } from 'recompose'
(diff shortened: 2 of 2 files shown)


src/packages/recompose/withState.js
```diff
@@ -16,12 +16,19 @@ export const withState = (
         : initialState
     }
 
+    hasUnmounted = false
```
istarkov: isMounted needs `componentDidMount` so it's better to check hasUnmounted

istarkov:
(_It looks like a week ago all projects required semicolons, and now all not, why?_)

chicoxyzzy:
codestyle was changed in https://github.com/acdlite/recompose/commit/5ab5c058224c6c6e093fe4377dafb7461040dad5

acdlite:
1. You can use a stub to test for console warnings, like so: https://github.com/acdlite/recompose/blob/master/src/packages/recompose/__tests__/compose-test.js#L30-L41

2. I'm not sure if it's our business to suppress this warning. This is normal React behavior, nothing specific to Recompose.

istarkov:
1) stub does not restore normal behavior if throw has occurred
2) May be it's normal React behavior but who wants this warning in the console? 
Almost every component with async callbacks use isMounted() check. And there is no way in functional component to check isMounted state and to avoid this error.

istarkov:
As an alternative:
`@isMounted` HOC which will add isMounted() check to props?

acdlite:
1. I don't know what you mean. You're just testing if `console.error` is called, right? Then when you're done, restore the stubbed method: https://github.com/acdlite/recompose/blob/master/src/packages/recompose/__tests__/compose-test.js#L40
2. I agree the warning is annoying for async callbacks, but I don't want to override built-in React behaviors.

An `isMounted()` helper is a good idea. I would probably accept a PR for that.

istarkov:
1) If expect in your example throw an error `expect(error.callCount).to.equal(0)` this code `console.error.restore()` will never be run, so all other tests get `console.error` redeclared.

2) isMounted got it

PR #57, opened 2015-11-13
Add isMounted HOC

istarkov:
isMounted helper HOC
https://github.com/acdlite/recompose/pull/52#issuecomment-156217124


acdlite:
Cool! Could we add an parameter that sets the name of the `isMounted` prop?

istarkov:
Done. 

Looks like lodash `curry` does not allow to use default arguments,

Like

```
const isMounted = (
  isMountedPropName='isMointed',
  BaseComponent
) =>
```

Because calling curried function without arguments returns function itself `curried == curried()`
So instead of `isMounted()` user needs to write `isMounted('isMounted')`

jimbolla:
React is planning on deprecating isMounted eventually. See https://github.com/facebook/react/issues/5465

istarkov:
This is not react isMounted call, see the source, using asynchronous calls we need to know does component still mounted before setState call, otherwise React gives an error message. 
So if you use `raf`, `promises`, other similar calls, there is no way without such method to avoid this error message.

istarkov:
`raf` is bad example above, we could cancel `raf` calls in `lifecycle` teardown hook,
but we can do nothing with promises.

istarkov:
Looks like it's not a problem to write cancelable support for promise,

like this 

```
const makeCancelable = (promise) => {
  let hasCanceled_ = false;

  return {
    promise: new Promise(
      (resolve, reject) => promise
        .then(r => hasCanceled_
          ? reject({isCanceled: true})
          : resolve(r)
        )
    ),
    cancel() {
      hasCanceled_ = true;
    },
  };
};

// HOW TO USE
const somePromise = new Promise(r => setTimeout(r, 1000));

const cancelable = makeCancelable(somePromise);

cancelable
  .promise
  .then(() => console.log('resolved'))
  .catch(({isCanceled, ...error}) => console.log('isCanceled', isCanceled));

// Cancel promise
cancelable.cancel();
```

so this problem can be solved using lifecycle HOC,
so I close this PR.

PR #65, opened 2015-11-18
Fix typo in withReducer example in README

happypoulp:
Hi,

Just a small typo fix in README regarding the use of withReducer.


acdlite:
Thanks!

PR #69, opened 2015-11-20
Remove uses of static class props

joshacheson:
Refactors uses of static class props to oldschool style for ie8 Compatibility.

Let me know how you feel about variable names.

### DIFF
--- a/src/packages/recompose/toClass.js
+++ b/src/packages/recompose/toClass.js
@@ -7,17 +7,18 @@ const toClass = baseComponent => {
     return baseComponent
   }
 
-  const NewComponent = class extends Component {
+  class ToClass extends Component {
     render() {
       return baseComponent(this.props, this.context)
     }
   }
-  NewComponent.displayName = getDisplayName(baseComponent)
-  NewComponent.propTypes = baseComponent.propTypes
-  NewComponent.contextTypes = baseComponent.contextTypes
-  NewComponent.defaultProps = baseComponent.defaultProps
 
-  return NewComponent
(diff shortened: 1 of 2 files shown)


src/packages/recompose/toClass.js
```diff
@@ -7,16 +7,17 @@ const toClass = baseComponent => {
     return baseComponent
   }
 
-  return class extends Component {
-    static displayName = getDisplayName(baseComponent)
-    static propTypes = baseComponent.propTypes
-    static contextTypes = baseComponent.contextTypes
-    static defaultProps = baseComponent.defaultProps
-
+  const NewComponent = class extends Component {
```
acdlite: The convention used in the other modules is for the class name to be the uppercase version of the helper. Also, use a normal class, not a class expression. So this would be:

```js
class ToClass extends Component {...}
```

src/packages/recompose/toClass.js
```diff
@@ -7,16 +7,17 @@ const toClass = baseComponent => {
...
     render() {
       return baseComponent(this.props, this.context)
     }
   }
+  NewComponent.displayName = getDisplayName(baseComponent)
```
acdlite: Small nitpick, but could you put a space between the closing bracket and the next line?

joshacheson:
@acdlite If you're cool with these changes, do you want me to do a rebase/squash?

acdlite:
Nah, I'm not a huge stickler about rebasing. Thanks!

acdlite:
Release in v0.11.2

PR #70, opened 2015-11-21
Improve rx-recompose example

Lendar:
I was playing with rx-recompose and got stuck when trying example. First, there was an error. Second, Counter was ignoring external props like `<Counter style={{color: 'red'}}/>`.

So, I had to patch it to:

- fix `ReferenceError: count is not defined`
- make mapPropsStream of example to actually pass props through

I'm just learning both Rx and recompose. Is my example correct? If not, I may re-submit PR tomorrow.

P.S.: Good job with rx-recompose :+1: 

UPDATED: do not merge yet, it's wrong :confused:

### DIFF
--- a/src/packages/rx-recompose/README.md
+++ b/src/packages/rx-recompose/README.md
@@ -76,7 +76,8 @@ const Counter = observeProps(
     return {
       increment: Observable.just(increment$),
       decrement: Observable.just(decrement$),
-      count
+      count: Observable.just(count$),
+      style: props$.pluck('style'),
     }
   },
   ({ count, decrement, increment, ...props }) => (


src/packages/rx-recompose/README.md
```diff
@@ -73,11 +73,16 @@ const Counter = observeProps(
       .startWith(0)
       .scan((count, n) => count + n)
 
-    return {
-      increment: Observable.just(increment$),
-      decrement: Observable.just(decrement$),
-      count
```
acdlite: Actually this could be fixed just by changing this line to `count: count$`

acdlite:
Good catch!

Here's the actual, working example from the test suite: https://github.com/acdlite/recompose/blob/master/src/packages/rx-recompose/__tests__/observeProps-test.js#L31-L42

If you update the PR to match that, I'll merge it. Otherwise I will fix it myself :)

Lendar:
@acdlite sorry to waste your time, but after I've re-checked your test case (which is not passing as it seems) and the examples I've noticed we were discussing the wrong branch (master).

Branch `rx-recompose` is fine. Both [tests](https://github.com/acdlite/recompose/blob/rx-recompose/src/packages/rx-recompose/__tests__/observeProps-test.js#L23-L27) and [example](https://github.com/acdlite/recompose/blob/rx-recompose/src/packages/rx-recompose/README.md) are using combineLatest so they are OK. Actually, npm page is OK as well :laughing:, so it's just me who went to docs at GitHub instead of npm.

PR #82, opened 2015-12-16
Fix variable name error

littlehaker:
```javascript
return {
      increment: Observable.just(increment$),
      decrement: Observable.just(decrement$),
      count: count$
}
```
instead of 
```javascript
return {
      increment: Observable.just(increment$),
      decrement: Observable.just(decrement$),
      count
}
```
in rx-recompose Readme demo.


acdlite:
Thanks!

PR #89, opened 2016-01-11
Allow `withState` to be required by ES5 code

dtinth:
Most functions in recompose can be required from ES5 code:

```js
const compose = require('recompose/compose')
```

But not `withState`.

```js
const withState = require('recompose/withState')
```

```js
withState('a', 'b', 0) // ERROR: withState is not a function
```

This is because `withState.js` exports more than one thing.


acdlite:
Good catch! Thanks!

acdlite:
Fix released in v0.14.4

PR #91, opened 2016-01-15
lifecycle: Allow `null` in place of `() => {}`

dtinth:
Some of our engineers passed `null` to `lifecycle()` instead of a no-op function `() => { }` or `_.noop`. This caused a run-time error in our staging environment.

This PR makes it allow falsy value to be a synonym for no-op. An alternative solution is to validate the arguments and throw immediately, so that these errors could be caught early.


acdlite:
Thanks!

acdlite:
Released in 0.14.5

PR #92, opened 2016-02-01
Update lodash to v4

Chrisui:
This PR simply updates the "lodash" dependency to v4 and fixes any references to old module locations or deprecated functions.

### DIFF
--- a/src/packages/recompose/mapPropsOnChange.js
+++ b/src/packages/recompose/mapPropsOnChange.js
@@ -1,6 +1,6 @@
 import { Component } from 'react'
-import pick from 'lodash/object/pick'
-import omit from 'lodash/object/omit'
+import pick from 'lodash/pick'
+import omit from 'lodash/omit'
 import shallowEqual from './shallowEqual'
 import createHelper from './createHelper'
 import createElement from './createElement'
--- a/docs/API.md
+++ b/docs/API.md
@@ -24,7 +24,7 @@ mapProps(
 
 Accepts a function that maps owner props to a new collection of props that are passed to the base component.
 
-`mapProps()` pairs well with functional utility libraries like [lodash-fp](https://github.com/lodash/lodash-fp). For example, Recompose does not come with a `omitProps()` function, but you can build one easily using lodash-fp's `omit()`:
+`mapProps()` pairs well with functional utility libraries like [lodash/fp](https://github.com/lodash/lodash/tree/npm/fp). For example, Recompose does not come with a `omitProps()` function, but you can build one easily using lodash-fp's `omit()`:
 
 ```js
 const omitProps = (keys, BaseComponent) => mapProps(omit(keys), BaseComponent);
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "karma-sinon": "^1.0.4",
     "karma-sourcemap-loader": "^0.3.5",
     "karma-webpack": "^1.7.0",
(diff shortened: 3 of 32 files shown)


Chrisui:
Trying to get tests to run locally to debug the `createHelper` issue but having some trouble.

```
Uncaught Error: Cannot find module "mock-modules"
  at /Users/chris/Projects/recompose/tests.webpack.js:148625 <- webpack:///src/packages/recompose-relay/~/react/~/envify/~/jstransform/src/__tests__/jstransform-test.js:19:0
```

Seems to be trying to test nested dependencies due to the configuration in `tests.webpack.js`. Any ideas how to prevent this?

Chrisui:
Ah just spotted https://github.com/acdlite/recompose/issues/51- will address later!

acdlite:
Hey thanks for doing this! I'll merge once the tests pass.

Chrisui:
Hey got these tests running locally (hopefully travis agrees!) but it seems there has been a change in expected lodash.curry functionality where `undefined` is handled as a parameter.

You can see the change in the latest commit here: https://github.com/acdlite/recompose/commit/07c1564fe7199ba92d130e5b479eb2148b3d0f79

acdlite:
Thanks!

billyjanitsch:
@Chrisui @acdlite FYI, 07c1564 breaks some valid previously-valid uses of recompose, e.g.:

``` js
import {withState} from 'recompose'

const MyComponent = props => <input {...props} />

withState('value', 'onChange', undefined, MyComponent)
// in 0.14.5, returns a valid component
// in 0.15.0, returns a curried function
```

Note that there's some discrepancy in how Lodash behaves -- see https://github.com/lodash/lodash/issues/1955 -- so the "new" behaviour might not have been intentional their part. Neglecting undefined arguments seems pretty surprising/non-standard to me.

acdlite:
@billyjanitsch You could pass `null` instead? But yeah this merits discussion. At the very least we should document the behavior and be explicit about supporting it.

billyjanitsch:
@acdlite Right, I replaced it with `null` when I figured it out. It's just a tricky bug to track down if you're already using `undefined`: it only crashes React during a batched render cycle (because `withState` *does* return a function -- just not a valid component function), so you get a pretty unhelpful error (no render-time stack trade, and no display name since the "component" is an anonymous partial.)

Btw, it looks like [this was a bug after all](https://github.com/lodash/lodash/issues/1955#issuecomment-181168289) and will be fixed in the next Lodash version. In the meantime, want a PR that reverts 07c1564 and replaces it with the suggested workaround?

acdlite:
Yeah, that sounds good. In fact, maybe we should just not use lodash's curry inside `createHelper()` at all, and instead always use our custom implementation? We could still wrap https://github.com/acdlite/recompose/blob/07c1564fe7199ba92d130e5b479eb2148b3d0f79/src/packages/recompose/createHelper.js#L31-L33 inside an environment check so it gets stripped out in production.

src/packages/recompose/mapPropsOnChange.js
```diff
@@ -1,6 +1,6 @@
 import { Component } from 'react'
-import pick from 'lodash/object/pick'
-import omit from 'lodash/object/omit'
+import pick from 'lodash/pick'
```
istarkov: Here we lost ability to pass function as depdendentPropKeys

PR #94, opened 2016-02-04
Fix missing import and export

### LINKED ISSUE #93: [rx-recompose] Observable is not defined when using rx-recompose 0.3.0-beta2

ahlee2326:
Resolves #93

### DIFF
--- a/src/packages/rx-recompose/createComponent.js
+++ b/src/packages/rx-recompose/createComponent.js
@@ -1,4 +1,5 @@
 import { Component } from 'react'
+import { Observable } from 'rxjs/Observable';
 
 const createComponent = propsToVdom =>
   class RxComponent extends Component {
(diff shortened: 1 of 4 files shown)


acdlite:
The idea is that rx-recompose should work with *any* Observable library, not just RxJS. So rather than import a specific library, we rely on Observable being globally polyfilled. This is the same approach most libraries use with respect to Promises.

ahlee2326:
But none of these polyfills are current supporting Observable (http://kangax.github.io/compat-table/esnext/), so I need to hack it by adding rxjs's Observable as a global variable like what your test does?

ahlee2326:
Nevertheless, this line is missing https://github.com/acdlite/recompose/pull/94/files#diff-a2ca7abe8afbfbe3e2af96bcad3b42d0R1

acdlite:
> But none of these polyfills are current supporting Observable (http://kangax.github.io/compat-table/esnext/), so I need to hack it by adding rxjs's Observable as a global variable like what your test does?

Yep! I'll add docs about this before it's released.

> Nevertheless, this line is missing

Oops. Good catch.

PR #95, opened 2016-02-09
Bump lodash and revert to correct curry behavior

billyjanitsch:
Re https://github.com/acdlite/recompose/pull/92#issuecomment-181163598:

Lodash 4.3.0 fixes the incorrect curry behavior. This bumps Lodash and reverts 07c1564.

I know that @acdlite proposed [not using Lodash's curry at all](https://github.com/acdlite/recompose/pull/92#issuecomment-181189168), but here's a simpler change in the meantime.


acdlite:
Thanks!

PR #96, opened 2016-02-10
Upgrade to Babel 6

billyjanitsch:
Also bump a few other deps while I'm at it.

Note that the curry test currently fails on master (and in this PR) because of the issue addressed in #95. Lodash bumped to 4.3 which fixes the behavior; this breaks the Recompose tests because the dependency is at ^4. Once #95 is merged and this PR is rebased, all tests should pass.

Also, the Babel dep is currently pinned at an exact version because of [this Relay bug](https://phabricator.babeljs.io/T7088) in the latest Babel.

Once this is merged, I'd like to add `babel-runtime` since Babel's injected functions are currently inflating the size of Recompose's build quite a bit.


acdlite:
Thanks for this work! I'll get this merged within the next few days.

Last time I tried it out, enabling babel-runtime reduced the minified bundle size *before* gzipping, but actually increased it *after* gzipping. Worth investigating again, though!

billyjanitsch:
@acdlite No problem! The Babel bug has been fixed, so I bumped to the latest version and rebased. The tests should pass now.

Re: babel-runtime, it's worth considering that even if it results in a larger Recompose bundle, it would allow any consumer of Recompose that also uses Babel (which I imagine most do) to share all of the babel-runtime code, thanks to npm's deduping.

acdlite:
Thank you! This was a big help

PR #100, opened 2016-02-24
Make createEventHandler accept an optional function for cancelling events

### LINKED ISSUE #80: createEventHandler should accept a function as an argument
When a FORM element submit event occurs, you need to return false and call event.preventDefault(). Otherwise the form will attempt to submit the form in the background before processing other event subscribers. 

The FuncSubject from https://github.com/fdecampredon/rx-react allows developers to define an immediate callback, which allows developers to prevent form submission. I'll can create a pull-request with the change.

Great work on this project, and your work in react-rx-component. Quite a beautiful way to build components.

dralletje:
Although I am not sure myself if this is a good idea, as it feels to me stuff like cancelling the event should be done outside of this (wrapping it in another function?!). Still, it wasn’t much effort to make this, just in case you do want to merge it ;) Fixes #80

### DIFF
--- a/src/packages/rx-recompose/README.md
+++ b/src/packages/rx-recompose/README.md
@@ -55,7 +55,9 @@ The second form is often more convenient because it avoids the need for `Observa
 ### `createEventHandler()`
 
 ```js
-createEventHandler(): Function & Subject
+createEventHandler(
+  callback: ?(value: any) => any
+): Function & Subject
 ```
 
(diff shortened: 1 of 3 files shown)


acdlite:
Yeah after thinking about this, not sure why it's needed after all. Do you really need to return false to prevent a form from submitting? You can just do `e.preventDefault()`, right?

dralletje:
And if not, you should just to:

```
<form
  onSubmit={e => {
    e.preventDefault()
    myHanderThing$()
    return false
  }}
/>
```

Issue #102, opened 2016-02-29
reverse props spread order for mapPropsOnChange


hartzis:
While working with `mapPropsOnChange` recently we noticed that `this.props` get spread over `computed props`.

https://github.com/acdlite/recompose/blob/master/src/packages/recompose/mapPropsOnChange.js#L24

It appears the original plan was to spread computed props last?
https://github.com/acdlite/recompose/issues/60#issuecomment-157454803

Example:
I create an `items` prop with `mapPropsOnChange`
```
mapPropsOnChange(['thing'], ({thing}) => ({items: thing.count()}))
```
but if `items` was already a prop being passed down the new computed prop of `items` will get overwritten by `this.props.items`.

I would expect my outputted computed props to be what gets passed down.

Let me know what you think, and I could get a PR in with tests this week.

istarkov:
For simplicity with line switch, omit on line 26 also can be removed. Otherwise it works like `mapAndFilterPropsOnChange`

hartzis:
@istarkov it appears `mapAndFilterPropsOnChange` is not a HOC in this library. I do like the idea of that though.

istarkov:
I meant that omit filtering was done because of merge order, if you switch lines 25 and 26, looks like there is no need to filter (omit) props in array.

hartzis:
I have the PR pretty much ready to go, but trying to run the tests local and they are currently failing with:
`Module build failed: Error: The (relay-query) Babel 5 plugin is being run with Babel 6.`

I think it is related to https://github.com/facebook/relay/pull/889, and when they publish a new version it may be fixed.

I may just submit the PR and see if the travis build fails.

PR #104, opened 2016-03-02
reverse spread of props on mapPropsOnChange

hartzis:
fix for #102


hartzis:
CI build failed due to:

Module build failed: Error: The (relay-query) Babel 5 plugin is being run with Babel 6.

I think it is related to https://github.com/facebook/relay/pull/889, and when they publish a new version it may be fixed.

hartzis:
blocked by tests failing, #106

hartzis:
@acdlite merged in master to get tests to run, and fixed the broken tests. Should be good to go now.

acdlite:
Thanks!

istarkov:
IMHO `omit` should be removed.

I'll explain why:

One of my most used construction with `mapPropsOnChange` was

```javascript
mapPropsOnChange(
  ['a', 'b', 'c'], 
  ({...props, a, b ,c}) => ({
    ...props,
    computedProps: blabla(a,b,c)
  })
),
```

spread operator allows me to not lost `a,b,c` in the future, and reuse them again.

Now to make this work I can't use spread operator at all, because other props out of selection array i.e. ['a', 'b', 'c'] will be placed in `computedProps` so will be outdated. 

The code above now looks like

```javascript
mapPropsOnChange(
  ['a', 'b', 'c'], 
  ({a, b ,c}) => ({
    a,
    b,
    c,
    computedProps: blabla(a,b,c)
  })
),
```

So every time I add property into selection array I need to add it into object to preserve. 

Removing `omit` will help a lot, I could just remove all that `...props`, and forget about property omitting logic in `mapPropsOnChange`

Issue #112, opened 2016-03-16
React 15.0.0-rc1 key warning 


zeroshine:
React 15.0.0-rc1 add a new warning.
- Elements will now warn when attempting to read ref and key from the props.  

recompose seems touch key from props, so this warning shows up

> `key` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop.

acdlite:
Thanks for the bug report. This can be fixed by using `hasOwnProperty` here: https://github.com/acdlite/recompose/blob/ca7c6d33de8afa4e48cee299ba9a16feba4b1d76/src/packages/recompose/createElement.js#L7

I'll fix soon!

acdlite:
Closed by #113. It'll go out in the next release.

PR #113, opened 2016-03-17
hasKey fix

fardormu:
fix for #112. Hopefully it's helpful.


acdlite:
Thanks! There's a babel-relay-plugin error that's causing the tests to fail, but I trust that this works.

Issue #66, opened 2015-11-18
expose `createElement` and `createHelper`


joshacheson:
These two private/util methods are really useful. It might be nice to either publish them as separate libraries on npm or expose them to consumers of recompose. Seems to me that people wanting to create their own helpers wont be a rare case, and these utilities are a huge help in that process.

acdlite:
I agree they could be useful to others. Currently they're available as

```js
import createElement from 'recompose/createElement'
import createHelper from 'recompose/createHelper'
```

but they should go on the main export, as well. Only roadblock right now is proper documentation.

acdlite:
`createElement` is now `createEagerElement` and is exposed (in master, not released yet). Now that currying has been removed, not sure there's any value in making `createHelper` public. I'll keep it private for now.

PR #116, opened 2016-03-20
expose and document createElement and createHelper

pwmckenna:
Addresses https://github.com/acdlite/recompose/issues/66.

Happy to update. It is certainly not perfect. There are a couple intentional omissions here:

- introduced `ReactChildrenType` but didn't really know how to concisely document the type of children...but then again, neither do the react docs ;)
- left out `createHelpers`'s `_helperLength`,  `setDisplayName` parameters. These don't seem to be used much even internally, and the leading underscore seems to indicate this should remain internal only, so I didn't document the optional params.


docs/API.md
```diff
@@ -556,3 +556,19 @@ hoistStatics(higherOrderComponent: HigherOrderComponent): HigherOrderComponent
...
 
 Augments a higher-order component so that when used, it copies static properties from the base component to the new component. This is helpful when using Recompose with libraries like Relay.
+
+### `createElement()`
+
+```js
+createElement(component: ReactElementType, props: Object, children: ReactChildrenType): ReactElementType
+```
+
+Creates a component similar to [`React.createElement`](https://facebook.github.io/react/docs/top-level-api.html#react.createelement), but safely handles assigning a key to a pure functional component by wrapping it in a class.
```
acdlite: I think you're confusing `createElement()` with `toClass()`. The purpose of the `createElement()` is to detect referentially transparent components and call them as functions rather than creating an unnecessary elements. See this doc for more details: https://github.com/acdlite/recompose/blob/master/docs/performance.md

pwmckenna: Fair enough, I'll update. I only really care about making `createHelper` public but it seemed rude to not take a crack at both.

acdlite: Haha, not rude at all.

acdlite:
Let's hold off on this until https://github.com/acdlite/recompose/issues/117 is resolved. In the meantime, you can import `createHelper` at `recompose/createHelper`.

Issue #60, opened 2015-11-17
mapPropsOnChange 


istarkov:
recompose does not allow to use most used pattern similar to `mapPropsOnChange`

What if I need to create computationally intensive work only if some props is changed, but pass other props as is. So I get all props online, and compute results of intensive tasks only when it needed.

For example I have property `x` and `data`, I need to make some intensive calculus if `data` changed but also I need to get current version of `x` in my component. With `mapPropsOnChange` I get the value of `x` on the moment `data` has changed. But every time I use this pattern I need new `x` values. 

This could be solved creating a HOC based on `mapPropsOnChange` but changing this line

https://github.com/acdlite/recompose/blob/master/src/packages/recompose/mapPropsOnChange.js#L23

on this

```javascript
return createElement(BaseComponent, {...this.props, ...this.childProps})
```

What are you think about
HOC with name and signature 

```
withPropsOnChange(
  depdendentPropKeys: Array<string>,
  propsMapper: (ownerProps: Object) => Object,
  BaseComponent: ReactElementType
): ReactElementType
```

acdlite:
The `propsMapper` parameter of `mapPropsOnChange()` receives the entire props object, not just the dependent props. So you can achieve what you're looking for like so:

```js
mapPropsOnChange(
  ['foo', 'bar'],
  ({ foo, bar, ...rest }) => ({
    foobar: computationallyIntensiveFunction(foo, bar) // computed prop
    ...rest // rest of props
  }),
  BaseComponent
)
```

istarkov:
This will not work!
propsMapper will be called only if `foo` or `bar` has changed,
but if changed something in the rest you will got the old childProps

istarkov:
As I wrote the only way to make it work is to combine props with childProps at render
```
return createElement(BaseComponent, {...this.props, ...this.childProps})
```

acdlite:
Oh... I see what you mean now. Need to write a failing test case.

istarkov:
But looks like mapPropsOnChange is the good name for current behavior, 
it looks like for new behavior it will be better to create something like `withPropsOnChange`?

acdlite:
Actually, probably need to rethink `mapPropsOnChange()` entirely. I don't like mapping the entire props object, but then mixing it back into the original props object.

istarkov:
Sounds good, thank you!

acdlite:
I dunno about `withPropsOnChange()` either... could work.

acdlite:
[...]

PR #119, opened 2016-03-26
removing omit

istarkov:
https://github.com/acdlite/recompose/pull/104#issuecomment-201820358

IMHO `omit` should be removed.

I'll explain why:

One of my most used construction with `mapPropsOnChange` was

```javascript
mapPropsOnChange(
  ['a', 'b', 'c'], 
  ({...props, a, b ,c}) => ({
    ...props,
    someResult: blabla(a,b,c)
  })
),
```

spread operator allows me to not lost `a,b,c` in the future, and reuse them again.

Now to make this work I can't use spread operator at all, because other props out of selection array i.e. ['a', 'b', 'c'] will be placed in `this.computedProps` so will be outdated. 

The code above after this [PR](https://github.com/acdlite/recompose/pull/104) now looks like

```javascript
mapPropsOnChange(
  ['a', 'b', 'c'], 
  ({a, b ,c}) => ({
    a,
    b,
    c,
    someResult: blabla(a,b,c)
  })
),
```

So every time I add property into selection array I need to add it into object to preserve. 

Removing `omit` will help a lot, I could just remove all that `...props`, and forget about property omitting logic in `mapPropsOnChange`

Also this simplify this component logic, I got a lot of questions from coworkers about this omitting functionality of `mapPropsOnChange`

### DIFF
--- a/src/packages/recompose/__tests__/mapPropsOnChange-test.js
+++ b/src/packages/recompose/__tests__/mapPropsOnChange-test.js
@@ -33,6 +33,8 @@ describe('mapPropsOnChange()', () => {
     renderIntoDocument(<StringConcat />)
 
     expect(omit(spy.getProps(), ['updateStrings'])).to.eql({
+      a: 'a',
+      b: 'b',
       c: 'c',
       foobar: 'ab',
       d: 'new'
@@ -44,6 +46,8 @@ describe('mapPropsOnChange()', () => {
     expect(mapSpy.callCount).to.equal(1)
 
     expect(omit(spy.getProps(), ['updateStrings', 'updateFoobar'])).to.eql({
+      a: 'a',
+      b: 'b',
       c: 'baz',
       foobar: 'ab',
       d: 'new'
@@ -51,6 +55,8 @@ describe('mapPropsOnChange()', () => {
 
     spy.getProps().updateStrings(strings => ({ ...strings, a: 'foo', 'b': 'bar', d: 'old' }))
     expect(omit(spy.getProps(), ['updateStrings', 'updateFoobar'])).to.eql({
+      a: 'foo',
+      b: 'bar',
       c: 'baz',
       foobar: 'foobar',
       d: 'new'
--- a/src/packages/recompose/mapPropsOnChange.js
+++ b/src/packages/recompose/mapPropsOnChange.js
@@ -1,6 +1,5 @@
 import { Component } from 'react'
 import pick from 'lodash/pick'
-import omit from 'lodash/omit'
 import shallowEqual from './shallowEqual'
 import createHelper from './createHelper'
 import createElement from './createElement'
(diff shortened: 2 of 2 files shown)


hartzis:
Apologies for not responding to your comment in #104. 

I'm all for this update. Thank you for putting this PR together.

I'm curious why the original `omit` was created? Maybe possibly making the assumption that since those props changed, you're computing something from them, and not going to need them after you've used them? 

But you can just create your own omit props with [mapProps](https://github.com/acdlite/recompose/blob/master/docs/API.md#mapprops).

istarkov:
Original `omit` was created after this issue https://github.com/acdlite/recompose/issues/60 it was a fast and working solution of that problem, 
with minimal changes to that version of `mapPropsOnChange`.

wuct:
The new behavior of `mapPropsOnChange` is closer to `withProps` rather than `mapProps`. Maybe we should rename it to `withPropsOnChange`?

istarkov:
@wuct you are right, withPropsOnChange really good name.
@acdlite what are you think about?

istarkov:
@acdlite please answer on this PR having this comment
https://github.com/acdlite/recompose/issues/124#issuecomment-207749189

As you see a lot of people and projects use this library, (I use it in every project now) and having React has changed I need to have a decision what to do this mapPropsOnChange now.

istarkov:
PS: I think it's not a big problem to write will or or not accept this, 
be sure I'm heavily depends on your decision (to use local HOC version or to wait with library updates)

acdlite:
I read your comment about `mapPropsOnChange` over here https://github.com/acdlite/recompose/issues/124#issuecomment-207749189 but haven't fully digested it yet.

My main concern is that it's not clear to me (or most people, it seems) what `mapPropsOnChange` is actually supposed to do. Maybe it's just a problem of documentation or naming, but I suspect it's a problem of not properly identifying the use case and designing the API around that. This at the top of my list to figure out before we cut a 1.0 release.

If you're pressed for time, I would recommend using your own local version until we sort this out. Thanks for your patience :)

istarkov:
Thank you, may be `withPropsOnChange` as supposed by @wuct will be the best choice. As yes mapPropsOnChange is not describe this component behavior.

acdlite:
Okay I finally took another look at this. I think you're right @istarkov — removing `omit` clears everything up, and I think the name `mapPropsOnChange` still works, too.

acdlite:
Merged! Thanks @istarkov!

I changed the behavior slightly... Instead of merging with the original props, it now acts like `mapProps()` and passes exactly what you return from the mapping function.

Will go out in the next release, soon.

hartzis:
@acdlite I'm pretty sure that removing the merge of computed props into the original props will break components that currently expect props that are now lost because of the spread removal.

I agree with and like this update, just wondering if this is going out in v1.0.0?

It is a simple change, just need to add the `...rest` in the function and then `{...rest, ...computedProps}` to the outputted props, but this change should be documented somewhere.

istarkov:
Changed bahavior will not work, as returning just `computedProps` you will get outdated results for `props` not in `shouldMapOrKeys`.

wuct:
@istarkov is right, I have created a PR to show what he said.

acdlite:
Blah you guys are right. Okay, let's add the merging back and rename it `withPropsOnChange()`

PR #120, opened 2016-03-27
depthEqual Proposal

dandelany:
**Proposal for a new `shouldComponent` helper called `depthEqual`**

I have a situation where I'd like to be able to efficiently pass nested object literals to a component, here's a contrived simple example:

```
<Point position={{x: this.state.pointX, y: this.state.pointY}} />
```

Ideally the `Point` component could have a `shouldComponentUpdate` method that would prevent re-rendering if `props.position.x` and `props.position.y` remained the same, even if the `props.position` reference changed. This is a bit difficult with the current API.

This PR adds a new function called `depthEqual` to the API, which is just like `shallowEqual` except that it allows you to pass a third argument `depth` which specifies how many nested object layers should be checked for reference equality. `depthEqual(a, b, 0)` is equivalent to `a === b`, and `depthEqual(a, b, 1)` is equivalent to `shallowEqual(a, b)`. The default depth argument is `1`.

This would allow the `Point` class to implement a `shouldComponentUpdate` like:

```
class Point extends React.Component {
  shouldComponentUpdate(nextProps) {
    return !depthEqual(this.props, nextProps, 2);
  }
}
```

Generally you would want to optimize performance by checking some props shallowly and some deeply, using something like `_.pick` and `_.omit`, but this is just a simple example. Hopefully you think this is useful - object literal props are an anti-pattern sometimes but I think they have legitimate use cases too.

A couple notes - I added a unit test but I couldn't get it to work using `import depthEqual from 'recompose'` like the other unit tests do. It seems like that syntax imports from the `npm install`ed version, how do you test your code in development? Do you `npm link` it to itself or something? I just used `../` syntax for now to get it to pass.

Also I updated `shallowEqual` to call `depthEqual` but left the existing unit tests in place, so there may be some duplication there. Feel free to clean some of these out if you think it's appropriate.

Thanks for taking a look!

-dan

### DIFF
--- a/src/packages/recompose/__tests__/depthEqual-test.js
+++ b/src/packages/recompose/__tests__/depthEqual-test.js
@@ -1,0 +1,130 @@
+import depthEqual from '../depthEqual'
+import { expect } from 'chai'
+
+// Adapted from https://github.com/rackt/react-redux/blob/master/test/utils/shallowEqual.spec.js
+describe('depthEqual()', () => {
+  it('returns true if arguments are equal', () => {
+    expect(depthEqual(expect, expect)).to.be.true
+    expect(depthEqual(expect, expect, 0)).to.be.true
+    expect(depthEqual(2, 2)).to.be.true
+    expect(depthEqual(2, 2, 0)).to.be.true
+  })
+
+  it('returns false if depth is 0 and !==', () => {
+    expect(
+      depthEqual(
+        { a: 1, b: 2, c: undefined },
+        { a: 1, b: 2, c: undefined },
+        0
+      )
+    ).to.be.false
+  })
+
+  it('returns true if depth is 1 and arguments fields are equal', () => {
+    expect(
+      depthEqual(
+        { a: 1, b: 2, c: undefined },
+        { a: 1, b: 2, c: undefined },
+        1
+      )
+    ).to.be.true
+
+    expect(
+      depthEqual(
+        { a: 1, b: 2, c: 3 },
+        { a: 1, b: 2, c: 3 }
+        // 1 should be default arg
+      )
+    ).to.be.true
+
+    const o = {}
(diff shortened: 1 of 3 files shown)


istarkov:
Thank you, as I know it is not well known equality check, and can cause some misunderstanding.

recompose allows you to make your checks without this addition
 - just use recompose `shouldUpdate` HOC to prevent update if point values has not changed

Also you could not change object reference if it properties has not changed - in most situation you could easily avoid your behavior on data level.

acdlite:
Thanks for the PR! Because, as @istarkov says, it's fairly easy to implement this type of helper using `shouldUpdate`, I would prefer not to add to the API.

Perhaps additional docs around `shouldUpdate` would be helpful, though.

Noitidart:
I wrote a depth equal using `shallowEqual` here - https://github.com/acdlite/recompose/issues/440

Noitidart:
I only ever needed depth of 0 or 1 so got rid of the above in favor of this:

```
export depth0Or1Equal(obj1, obj2, depth1Keys={}) {
    // depth1Keys is objecect of keys you want to test at depth of 1, by default test depth of 0. value can be anything.
    const keys1 = Object.keys(obj1);
    const keys2 = Object.keys(obj2);
    if (keys1.length !== keys2.length) return false;

    for (const key of keys1) {
        if (!(key in obj2)) return false;

        const val1 = obj1[key];
        const val2 = obj2[key];
        if (key in depth1Keys) {
            if (!shallowEqual(val1, val2)) return false;
        } else {
            if (val1 !== val2) return false;
        }
    }

    return true;

}
```

Usage:

```
depth0Or1Equal({a:[]}, {a:[]}) === false
depth0Or1Equal({a:[]}, {a:[]}, {a:1}) === true
```

PR #123, opened 2016-04-06
add suport for previous pops on props change

matanwerbner:
this caused a problem for me when using recompose - i had to have access to existing props on the callback, and they were'nt there...

### DIFF
--- a/src/packages/recompose/doOnReceiveProps.js
+++ b/src/packages/recompose/doOnReceiveProps.js
@@ -5,11 +5,11 @@ import createElement from './createElement'
 const doOnReceiveProps = callback => BaseComponent =>
   class extends Component {
     componentWillMount() {
-      callback(this.props)
+      callback(this.props, null)
     }
 
     componentWillReceiveProps(nextProps) {
-      callback(nextProps)
+      callback(nextProps, this.props)
     }
 
     render() {


istarkov:
Agree with this PR. (Except `null`, I prefer `undefined` as it allows to use default values)

BTW I've never used this HOC, 
in local projects I use modified version of this HOC.

With next additions:
1) In most cases I need to do something in `componentWillReceiveProps` if something has changed, so I use `shallowEqual` on selected props. Like

```javascript
doOnPropsChange(
  ['someProp', 'someOtherProp'],
  (props) => { do something }
```

2) The next problem is state change, for example

```javascript
withState('someState', 'setSomeState', blabla),
doOnPropsChange(
  ['someProp', 'someOtherProp'],
  ({ setSomeState }) => { 
    setSomeState(blublu);
  }
)
```

As state is coming from parent this causes additional render. In some of my cases this causes desynchronization of overall component state. So I added ability to prevent first render.

```javascript
withState('someState', 'setSomeState', blabla),
doOnPropsChange(
  ['someProp', 'someOtherProp'],
  ({ setSomeState }) => { 
    setSomeState(blublu);
    return PREVENT_RENDER_ONCE; // exported symbol
  }
)
```

The problem that this `PREVENT_RENDER_ONCE` should be called explicitly as only user know has the state changed or not.

I can't say that I like my solution, It can give unpredictable result with `PREVENT_RENDER_ONCE` logic, I feel it can be more easier, so let's discuss what we can do with current HOC. 


---

My local version

```javascript
import { Component } from 'react';
import pick from 'lodash/pick';
import shallowEqual from 'recompose/shallowEqual';
import createHelper from 'recompose/createHelper';
import createElement from 'recompose/createElement';

export const PREVENT_RENDER_ONCE = Symbol('Prevent redraw on current call');

const doOnPropsChange = (depdendentPropKeys, callback, BaseComponent) => {
  const pickDependentProps = props => pick(props, depdendentPropKeys);

  return class extends Component {
    lastCall_ = undefined;

    componentWillReceiveProps(nextProps) {
      if (!shallowEqual(
        pickDependentProps(this.props),
        pickDependentProps(nextProps)
      )) {
        this.lastCall_ = callback(nextProps, this.props);
      }
    }

    shouldComponentUpdate() {
      / / A BIG PROBLEM: If callback returns PREVENT_RENDER_ONCE but where were no setState this shows old result
      if (this.lastCall_ === PREVENT_RENDER_ONCE) { 
        this.lastCall_ = undefined;
        return false;
      }

      return true;
    }

    render() {
      return createElement(BaseComponent, this.props);
    }
  };
};

export default createHelper(doOnPropsChange, 'doOnPropsChange');
```

acdlite:
Like @istarkov said, we're probably going to remove this from the next version, and point people to either rx-recompose or normal React classes, instead.

acdlite:
Thanks for the PR, though! Closing for the above reasons

PR #125, opened 2016-04-07
Allow react@15 as a peer dependency

billyjanitsch:
React 15 [was released today](https://facebook.github.io/react/blog/2016/04/07/react-v15.html). All of the tests for recompose & friends still pass, so I think it should be safe to update the peer dep. (Things are also working fine in our build.)


package.json
```diff
@@ -51,9 +51,9 @@
     "mocha": "^2.3.3",
     "phantomjs-prebuilt": "^2.1.3",
     "pretty-bytes": "^3.0.1",
-    "react": "^0.14.2",
-    "react-addons-test-utils": "^0.14.2",
-    "react-dom": "^0.14.2",
+    "react": "^0.14.2 || ^15.0.0",
```
istarkov: not a good idea to write ` || ` version in devDependency,
It's normal for peerDeps, but devDeps should be more explicit

billyjanitsch: Good point, thanks for the catch. Fixed!

acdlite:
Thanks!

Issue #85, opened 2015-12-20
Replace `createSpy()` with proper testing library


acdlite:
`createSpy()` is a temporary hack I came up with until a proper solution for testing function components emerged. I would really like to get rid of it.

Once there is stable, widely-used testing framework that supports functional components, let's remove `createSpy()`. Candidates:

- **[enzyme](https://github.com/airbnb/enzyme)** — Maintained by Airbnb. Currently doesn't work in browser environments (e.g. mocha), which could be a dealbreaker
- **[teaspoon](https://github.com/jquense/teaspoon)** — Works really well in my limited testing. Leaning toward this right now.

PR #126, opened 2016-04-08
withHandlers + docs

acdlite:
Adds a new helper `withHandlers`, which replaces `withAttachedProps`. Also includes updates to the API docs and README that I forgot to push when automatic currying was removed.


acdlite:
Also used Enzyme for the tests, which worked out quite nicely. Should migrate the rest of the tests at some point (https://github.com/acdlite/recompose/issues/85).

src/packages/recompose/withHandlers.js
```diff
@@ -0,0 +1,35 @@
+import { Component } from 'react'
+import mapValues from 'lodash/mapValues'
+import createElement from 'recompose/createElement'
+import createHelper from 'recompose/createHelper'
+
+const withHandlers = handlers => BaseComponent =>
+  class extends Component {
```
istarkov: Could `this.handlers` could be memoized as `handlers` never change, so you could convert this into functional component, without need of `extends Component`

docs/API.md
```diff
@@ -1,43 +1,64 @@
...
 mapPropsOnChange(
   depdendentPropKeys: Array<string>,
-  propsMapper: (dependentProps: Object) => Object,
-  BaseComponent: ReactElementType
-): ReactElementType
+  propsMapper: (dependentProps: Object) => Object
+): HigherOrderComponent
 ```
 
 Similar to as `mapProps()`, but child props are only re-computed when one of the props specified by `dependentPropKeys` has changed. This helps ensure that computationally intense `propsMapper` functions are only executed when necessary.
```
nfcampos: typo: 'Similar to ~~as~~'

docs/API.md
```diff
@@ -46,48 +67,74 @@ Similar to as `mapProps()`, but child props are only re-computed when one of the
...
+```js
+const enhanceForm = compose(
+  withState('value', 'updateValue', ''),
+  withHandlers({
+    onChange: props => event => {
+      props.updateValue(event.target.value)
+    },
+    onSubmit: props => event => {
+      event.preventDefault()
+      submitForm(props.value)
```
nfcampos: I don't think this example works... From the implementation of `withHandlers` it looks like the handlers aren't re-created when props change so that `onSubmit` will call `submitForm` with the initial `props.value`, no?

src/packages/recompose/withHandlers.js
```diff
@@ -0,0 +1,35 @@
+import { Component } from 'react'
+import mapValues from 'lodash/mapValues'
+import createElement from 'recompose/createElement'
+import createHelper from 'recompose/createHelper'
+
+const withHandlers = handlers => BaseComponent =>
```
istarkov: I meant something like this
```
const withHandlers = handlers => BaseComponent => ((() => {
  let lastProps_ = undefined;

  const newHandlers = mapValues(
    handlers,
    createHandler => (...args) => {
      const handler = createHandler(lastProps_);
      return handler(...args);
    }
  );
  // FUNCTIONAL !!!
  return (props) => {
    lastProps_ = props;

    return createElement(
      BaseComponent,
      {
        ...props,
        ...newHandlers,
      }
    );
  };
})());
```

docs/API.md
acdlite: See the test file, it's essentially the same thing as that. On my phone walking home but I'll elaborate later.

nfcampos: On the test file you only have the `onChange` handler, which works since the only prop it uses is the `updateValue` function, which never changes. Unless I'm missing something here the onSubmit

acdlite: It's may be hard to tell from the source, but each handler prop that is passed to the base component essentially looks like this:

```js
const handler = (...args) => createHandler(this.props)(...args)
```

It closes over the current value of the props.

You're correct that the test doesn't actually test this, though. I'll add another one to prove it works.

acdlite: See the updated test :)

nfcampos: Ahh, now I get it :) that error checking in the middle threw me off. This looks really useful!

mrtnbroder:
I need dis. Is it ready? :D

acdlite:
Will try to get it merged and released sometime today

acdlite:
Released in 0.17

PR #129, opened 2016-04-15
Use enzyme for component tests

acdlite:
- [x] Use enzyme for component tests, instead of React TestUtils + the hacky `createSpy` workaround we were using. Closes https://github.com/acdlite/recompose/issues/85.
- [x] Use Ava as a test runner, instead of Mocha.

Ava runs each test file in a separate process. Travis will kill processes after a certain number. We're using a workaround until Ava provides a way to limit the number of processes. https://github.com/sindresorhus/ava/issues/718


briandipalma:
Just curious did moving to Ava speed the test runs up?

amccloud:
I'm also curious if Ava is a speed up here. Ava for me was about 4x slower because the overhead of spinning up each process.

PR #132, opened 2016-04-21
Add a test for mapPropsOnChange

wuct:
This PR shows an unexpected behavior of `mapPropsOnChange`.

Please see #119


wuct:
I am not sure why the Travis CI build passed, but there is a test failed (as expected).

```
  ✖ mapPropsOnChange maps subset of owner props to child props   
  t.is(div.prop('c'), 'baz')
           |                
           "c"              
  
  1 test failed
  1. mapPropsOnChange maps subset of owner props to child props
  AssertionError: 'c' === 'baz'
```

acdlite:
@wuct Yeah I'm aware of the false positives on Travis. It's due to the Ava workaround script we're using.

PR #133, opened 2016-04-21
Use codecov instead of codecov.io

wuct:
[`codecov.io`](https://github.com/cainus/codecov.io) is deprecating, so it might be better to switch to [`codecov`](https://github.com/codecov/codecov-node).


acdlite:
Thanks!

Issue #105, opened 2016-03-03
`withProps` documentation + potential fix


npbee:
Hello and a quick thanks for the library!  I've been enjoying it a lot.

I noticed that the `withProps` function is documented as only accepting an object, but seems to be written to also accept a function:

**Docs:**
### `withProps()`

```js
withProps(
  props: Object,
  BaseComponent: ReactElementType
): ReactElementType
```

**Src:** 
```js
const withProps = (input, BaseComponent) => {
  let getProps
  const props = isFunction(input)
    ? input(getProps)
    : input

  return ownerProps => {
    getProps = () => ownerProps
    return createElement(BaseComponent, {
      ...ownerProps,
      ...props
    })
  }
}
```

Secondly, I don't think the function form will work as expected because at the time of the function call, `getProps` is still `undefined`.  If we were to move it below the `getProps` definition I think it'd work:

```js
const withProps = (input, BaseComponent) => {
  let getProps
  let props

  return ownerProps => {
    getProps = () => ownerProps
    props = isFunction(input)
      ? input(getProps) 
      : input
    return createElement(BaseComponent, {
      ...ownerProps,
      ...props
    })
  }
}
```

I'd be happy to open up a PR with some fixes but just wanted to check first to make sure it was the intended idea.

wuct:
Thanks @npbee, I did not notice that `withProps` accepts function.
Regarding there is no difference between `withAttachedProps` and `withProps` with a function as first parameter, I would like to propose another implementation.

```
const withProps = (input, BaseComponent) => {
  return ownerProps => {
    props = isFunction(input)
      ? input(ownerProps) 
      : input
    return createElement(BaseComponent, {
      ...ownerProps,
      ...props
    })
  }
}
```

We can use it like `mapProps` but more concisely in some cases.

```
mapProps(props => ({
  ...props,
  foo: doSomething(props)
}))
withProps(props => ({ foo: doSomething(props) }))
```

However, I never use `withProps` currently. In my opinion, `mapProps` and `withAttachedProps` are enough to replace `withProps` totally.

npbee:
Hi @wuct, good points!  In fact I ended up using `mapProps` in the same way you suggested and it worked fine.

[...]

PR #137, opened 2016-04-23
Fix withProps #105

istarkov:
Fix this https://github.com/acdlite/recompose/issues/105#issuecomment-213607329


wuct:
@istarkov thanks for your work. This version of `withProps` is like what I proposed in #105 but much more concise, good job!

acdlite:
👍

clessg:
Thanks, @istarkov!

Any idea when this might get published?

PR #141, opened 2016-04-27
Pull in API changes from next branch

acdlite:
The `next` branch has some nice improvements, but they won't be released until RxJS 5 is finalized. In the meantime, we should port those improvements over to master.

- [x] Add `createComponent`
- [x] `createEventHandler` returns a separate stream and handler, like Elm mailboxes.
- [x] Remove support for object-of-streams return value.
- [x] Rename `observeProps` to `mapPropsStream` (not part of `next` branch, but let's go ahead and do this now, anyway).

### DIFF
--- a/src/packages/rx-recompose/createEventHandler.js
+++ b/src/packages/rx-recompose/createEventHandler.js
@@ -1,21 +1,18 @@
-import { Subject } from 'rx'
+import { Observable } from 'rx'
 
-// Idea and implementation borrowed from
-// https://github.com/fdecampredon/rx-react
 const createEventHandler = () => {
-  function subject(value) {
-    subject.onNext(value)
+  const observers = []
+  const stream = Observable.create(observer => {
+    observers.push(observer)
+    return () => {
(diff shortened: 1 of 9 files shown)


src/packages/rx-recompose/createEventHandler.js
```diff
@@ -1,21 +1,18 @@
-import { Subject } from 'rx'
+import { Observable } from 'rx'
 
-// Idea and implementation borrowed from
-// https://github.com/fdecampredon/rx-react
 const createEventHandler = () => {
-  function subject(value) {
-    subject.onNext(value)
+  const observers = []
```
istarkov: Just curious, isn't it will be the same just creating single observer, and exporting stream as `stream.publish()` ?

acdlite: I basically copy-pasted this from the `rx-recompose-next` branch, where I want to avoid a dependency on RxJS. If you can think of a better way that doesn't add a dep, that'd be sweet :)

Issue #146, opened 2016-04-28
createElement optimisation


nfcampos:
if the signature of `createElement` was changed to `(Component) => (props, children) => ` then we could incur the cost of choosing the appropriate branch only once. This is made possible by the fact that for most uses (all `recompose` internal uses, at least) there's no use in checking for `key` in `props` as props passed to a component have `key` removed by `react`.
`createElement` would become something like 
```js
const createElement = (Component) => {
  if (isReferentiallyTransparentFunctionComponent(Component)) {
    const component = Component
    return (props, children) => {
      if (children) {
        return component({ ...props, children })
      }
      return component(props)
    }
  }

  return (props, children) => {
    if (children) {
      return <Component {...props}>{children}</Component>
    }

    return <Component {...props} />
  }
}
```

and users of `createElement`, eg. `shouldUpdate` something like this
```js
const shouldUpdate = test => BaseComponent => {
  const create = createElement(Component)
  return class extends Component {
    shouldComponentUpdate(nextProps) {
      return test(this.props, nextProps)
    }

    render() {
      return create(this.props)
    }
  }
}
```
What do you think?

acdlite:
Neat idea, but couldn't we achieve the same thing with

```js
const shouldUpdate = test => BaseComponent => {
  const create = createElement.bind(null, Component)
  return class extends Component {
    shouldComponentUpdate(nextProps) {
      return test(this.props, nextProps)
    }

    render() {
      return create(this.props)
    }
  }
}
```

?

Or we could curry `createElement` so that both forms work. I think there's value in `createElement` being a drop-in replacement for `React.createElement()`.

acdlite:
Oh duh obviously `.bind` won't work haha. Then I would prefer a "curried" version that supported both signatures.

acdlite:
Actually, I don't think there's a way to "curry" `createElement` without making it incompatible with React's version. Maybe we need a separate version that we can use internally.

acdlite:
Closed by #147

PR #147, opened 2016-04-28
Use "curried" form of createElement internally

acdlite:
See #146.

@nfcampos Want to review?

### DIFF
--- a/src/packages/recompose/createElement.js
+++ b/src/packages/recompose/createElement.js
@@ -2,12 +2,12 @@ import React from 'react'
 import isReferentiallyTransparentFunctionComponent
   from './isReferentiallyTransparentFunctionComponent'
 
-const createElement = (Component, props, children) => {
+const _createElement = (isReferentiallyTransparent, Component, props, children) => {
   /* eslint-disable */
   const hasKey = props && props.hasOwnProperty('key')
   /* eslint-enable */
 
-  if (!hasKey && isReferentiallyTransparentFunctionComponent(Component)) {
+  if (!hasKey && isReferentiallyTransparent) {
     const component = Component
     if (children) {
       return component({ ...props, children })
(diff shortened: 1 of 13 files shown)


src/packages/recompose/createElement.js
```diff
@@ -2,12 +2,12 @@ import React from 'react'
...
   from './isReferentiallyTransparentFunctionComponent'
 
-const createElement = (Component, props, children) => {
+const _createElement = (isReferentiallyTransparent, Component, props, children) => {
   /* eslint-disable */
   const hasKey = props && props.hasOwnProperty('key')
   /* eslint-enable */
 
-  if (!hasKey && isReferentiallyTransparentFunctionComponent(Component)) {
+  if (!hasKey && isReferentiallyTransparent) {
```
nfcampos: for internal use there's no point in checking `hasKey` since react removes `key` from props before passing them down to the component https://github.com/facebook/react/blob/master/src/isomorphic/classic/element/ReactElement.js#L149)

nfcampos:
other than the `key` thing this is exactly what I had in mind 👍

acdlite:
The reason for the check is that elements with keys aren't referentially transparent:

```js
// Given an otherwise referentially transparent component
const Item = ({ item }) =>
  <li>{item.title}</li>

const List = ({ items }) =>
  items.map(item => <Item key={item.id} item={item} />)

// NOT the same as
const List = ({ items }) =>
  items.map(item => Item({ key: item.id, item }))
```

The `key` is lost and React will print a warning.

nfcampos:
I know, but the internal uses of `createElement` will never fall into that situation because we never assign a `key` in recompose and the props all react elements receive from their parent never have a key. Even if the parent provided a `key` `React.createElement` removes it before passing the props down to the component.
ie.
```js
const Item = ({ item, key }) =>
  // key will be undefined
  <li>{item.title}</li>

const List = ({ items }) =>
  items.map(item => <Item key={item.id} item={item} />)
```

acdlite:
For internal purposes, you're right, we don't ever use keys. I'll update the PR. I'd still like the "public" version to be a drop-in replacement for `React.createElement`.

acdlite:
@nfcampos See latest commit :)

nfcampos:
Looks great :)

PR #151, opened 2016-05-04
Allow for a version of 'pure' that also checks state parity

toddtarsi:
It's a bad time if your pure component has state changes and your props don't.
Perhaps I'm doing something conceptually wrong, but I figured I'd share this. Thanks for the great lib!


istarkov:
Thank you for your contribution!
 
Where is no state at this HOCs, (there are no calls to `setState` etc) so there is no need to compare state.

toddtarsi:
@istarkov - Thanks for the quick response! Got it, your library is specifically designed for stateless HOCs. I'll use pure on our stateless components, and a small utility function for our stateful components.

Feel free to close at your leisure. :)

istarkov:
Even for statefull components, pure HOC is a component itself, so also can't have state ;-)
In `recompose` you can use `withState` HOC for state, and as withState state is passed down as props, just `pure` on props is enough.

acdlite:
To try to clarify it further: `pure` will work for any component, even stateful ones. But, if you use it on a stateful component, it will only prevent updates that are triggered by new props, not updates that are triggered by state.

wuct:
In addition, if you like the decorator flavor, you can use [pure-render-decorator](https://github.com/felixgirault/pure-render-decorator).

PR #159, opened 2016-05-14
brought back doOnReceiveProps() with example and better API

belkosyak:


acdlite:
Thanks for the PR, but `doOnReceiveProps()` was removed because it encouraged anti-patterns. Instead, https://github.com/acdlite/recompose/pull/162 provides an updated `lifecycle()` helper that is a thin wrapper around `React.createClass()`. This is the only lifecycle-related helper we'll provide. (At least for now—it's possible we'll reconsider in the future.)

belkosyak:
Yep, `lifecycle()` looks cool! We will use it to create a HOC like `doOnReceiveProps()` then

Issue #68, opened 2015-11-19
add mount option to lifecycle


hartzis:
```
lifecycle(
  setup: (component: ReactComponent) => void,
  mount: (component: ReactComponent) => void,
  teardown: (component: ReactComponent) => void,
  BaseComponent: ReactElementType
): ReactElementType
```

Add `mount` option to `lifecycle` that calls your function during `componentDidMount`. This will allow use for isomorphic/universal apps when you only want things ran on client and not on server.

Let me know what you think. I could probably get a PR in this week.

jeron-diovis:
+1 for this issue.
It looks inconsistent for now. I.e., we already dealing with DOM lifecycle hooks, in `willUnmount` - so why there is no other DOM hooks?

acdlite:
I've been thinking lately about removing `lifecycle` all together. The lifecycle methods are essentially useless without access to the component instance, in which case you should just use a class.

acdlite:
Does this make sense to anyone else? Is `lifecycle` really all that useful?

jeron-diovis:
> The lifecycle methods are essentially useless without access to the component instance
Well, not exactly. For example, you can start some timers when component is mounted - note, not in constructor - as it will cause problems with server-side rendering. Or, my own case, with which I came to this issue - a print page. It looks like this:
```
class PrintPage extends React.Component {
  render() {
    return (
      <Content {...this.props} />
    );
  }

  componentDidMount() {
    window.print();
  }
}
```
With `didMount` hook it could be just `lifecycle(window.print, Content)`.

Sure, it's all a very edge cases. But they exist, and it's always nicely when there is a handy function which saves you from manually writing another boilerplate class =)

istarkov:
Have a question about:

What the idea to pass `this` as an argument to lifecycle methods, isn't it better to pass just props and some internal object to hold data across lifecycle events (_like window event handlers or some instance methods which depends on lifecycle_).

To get state access user can use `lifecycle` in combination with `withState`. 

Does something like this will not be enough?

```javascript
 class Lifecycle extends React.Component {
   internalObj_ = null
 
   componentWillMount() {
       this.internalObj_ = setup(props)
    }

    ...otherLifeCycleEventsIfNeeded

    componentWillUnmount() {
       teardown(this.props, this.internalObj_)
    }

    render() {
      return createElement(BaseComponent, {...this.props, ...this.internalObj_})
    }
  }
```

istarkov:
[...]

PR #162, opened 2016-05-15
Add back lifecycle() as wrapper around React.createClass

### LINKED ISSUE #156: Request to bring back lifecycle
Hi, wanted to know the reason why you removed the `lifecycle` helper on your latest release (v0.18.0) and maybe open a vote on bringing it back in some form if anyone else uses / requires this feature.

My use case looks sth like this:

```js
const options = new Options({})

export const mapOptions = (options) => {
	return compose(
		withState('options', 'updateOptions', options.get()),
		mapProps(({updateOptions, ...rest}) => ({
			onUpdateOptions: () => {
				updateOptions(() => options.get())
			},
			...rest,
		})),
		lifecycle((component) => {
			options.on('update', component.props.onUpdateFreezer)
		}, () => {
			options.off('update', component.props.onUpdateFreezer)
		})
	)
}

const options = new Options({})
const MyOptionComponent = mapOptions(options)(MyComponent)
options.set({foo: 'bar'}) // triggers 'update'
```

Basically `options` triggers `update` event on which I would like to update the props. Maybe there's a other (better) way to do this without creating a base class and extending it!?

Cheers

acdlite:
Closes #156 and #68

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -407,6 +407,16 @@ getContext(
 
 Gets values from context and passes them along as props. Use along with `withContext()`.
 
+### `lifecycle()`
+
+```js
+lifecycle(
+  spec: Object,
+): HigherOrderComponent
+```
+
+A higher-order component version of [`React.createClass()`](https://facebook.github.io/react/docs/top-level-api.html#react.createclass). It supports the entire `createClass()` API, except the `render()` method, which is implemented by default (and overridden if specified; an error will be logged to the console). You should use this helper as an escape hatch, in case you need to access component lifecycle methods.
+
 ### `toClass()`
 
 ```js
--- a/src/packages/recompose/__tests__/lifecycle-test.js
+++ b/src/packages/recompose/__tests__/lifecycle-test.js
@@ -1,0 +1,17 @@
+import test from 'ava'
+import React from 'react'
+import { mount } from 'enzyme'
+import { lifecycle } from '../'
+
+test('lifecycle is a higher-order component version of React.createClass', t => {
+  const enhance = lifecycle({
+    componentWillMount() {
+      this.setState({ bar: 'baz' })
+    }
+  })
+  const Div = enhance('div')
+
+  const div = mount(<Div foo="bar" />).find('div')
+  t.is(div.prop('foo'), 'bar')
+  t.is(div.prop('bar'), 'baz')
+})
--- a/src/packages/recompose/__tests__/lifecycle-test.js
+++ b/src/packages/recompose/__tests__/lifecycle-test.js
@@ -10,6 +10,7 @@ test('lifecycle is a higher-order component version of React.createClass', t => 
     }
   })
   const Div = enhance('div')
+  t.is(Div.displayName, 'lifecycle(div)')
 
   const div = mount(<Div foo="bar" />).find('div')
   t.is(div.prop('foo'), 'bar')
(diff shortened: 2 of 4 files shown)


dlebedynskyi:
Note you now must use `function` instead of arrow functions 

```
lifecycle({
  	componentDidMount: function() {
     console.log('props', this.props);
    },
    componentWillReceiveProps: function (newProps) {
     console.log('new props', newProps);
     console.log('props', this.props);
    }
  }),
```
since *props are bound to context of HoC*. 
@acdlite  you might want to put it into documentation since `doOnReceiveProps`, as one of methods  it replaced,  was working differently. 
Also I update fiddle to show lifecycle https://jsfiddle.net/v6jojyej/2/

erlis:
@acdlite, 

I was checking the jsfiddle shared by @dlebedynskyi and it really helped me in understanding what the method does. 

The documentation for recompose is really bad, most of the time when I need to understand how something works I have to open a test and read through the code because the description on the API page doesn't help much. 

I think adding some code example to the API page will help tremendously, also I like the jsfiddle idea because I can play with the example.

acdlite:
I agree the documentation sucks. Gotta fix it soon.

dlebedynskyi:
@acdlite as an proposal on lifecycle 
instead of method as a normal unbind function how about to use the same approach for lifecycle methods as in withHandlers 
something like this 

```
lifecycle: { 
componentWillMount: (props) => (..args) => { ... },
componentWillReceiveProps: (props) => (newProps) => {... } 
}
```
in this case we will avoid problem like above, make code styles consistent across all recompose. And overall we really just use simple  function to have access to `this.props`. What do you think?

wuct:
I think the current implementation is fine. It is easy to understand and friendly to people who are new to `recompose`. We just need to improve the doc.

The new approach will lose the ability to access `this`, which I sometimes find useful.

dlebedynskyi:
@wuct valid point too. we just pretty much never use this due to recompose. 
And I think in context this will be HoC lifecycle one, correct? 
how do you think that will be used?

bholloway:
(Please excuse me in advance if I am misunderstanding the situation)

@wuct accessing `this` breaks encapsulation on the wrapped component. I think it is good if `recompose` helps you better compose your classes but I don't think this is the majority use case. Right now it feels like `lifecycle` is from a different library.

At face value I like @dlebedynskyi's proposal. I sense there is no progress because existing users will be disadvantaged, which is a fair concern.

In some ways I think it is down to a bad choice of names - `lifecycle` to me implies we are dealing with the props lifecycle (`..WillMount`, `..WillRecieveProps`, `..WillUnmount`). In comparison the current `lifecycle` implementation is just a mixin.

In hindsight it would be better called `mixin`, which serves @wuct's use case. And `lifecycle` would be free to implement something like @dlebedynskyi is suggesting.

Since this is not the case we just need to find another good name for @dlebedynskyi 's component in order to move forward.

wuct:
@dlebedynskyi @bholloway Yes, to create/teardown info (ex: event listeners) in the lifecycle is one of the cases. Another case I encounter is to find the DOM element and get the layout info, like `offsetLeft`, `clientWidth`, etc.

wuct:
@bholloway and yes, the current implementation of `lifecycle` is just a `mixin`. Renaming might be helpful.

PR #163, opened 2016-05-15
rx-recompose: createEventHandler: add stream initializer function

jeron-diovis:
From docs:
```js
  const { handler: increment, stream: increment$ } = createEventHandler()
  const { handler: decrement, stream: decrement$ } = createEventHandler()
  const count$ = Observable.merge(
      increment$.map(() => 1),
      decrement$.map(() => -1)
    )
```

Logically, `increment$` and `decrement$` should already contain +1/-1 streams. 
Additionally, when you need to use that streams more than once, it turns into smth like this:
```js
const { handler: decrement, stream: _decrement$ } = createEventHandler()
const decrement$ = _decrement$.map(() => -1)
doStuff1(decrement$)
doStuff2(decrement$)
...
```

So, I think, it would be nice to replace it with this:
```js
const { handler: increment, stream: increment$ } = createEventHandler($ => $.map(() => 1))
const { handler: decrement, stream: decrement$ } = createEventHandler($ => $.map(() => -1))
const count$ = Observable.merge(increment$, decrement$)
```


acdlite:
Hmm I like this, but won't almost all uses of this initializer function just be a call to `map()`? Maybe the initializer should be the projection function passed to map instead:

```js
const { handler: increment, stream: increment$ } = createEventHandler(() => 1))
const { handler: decrement, stream: decrement$ } = createEventHandler(() => -1))
const count$ = Observable.merge(increment$, decrement$)
```

For anything more complicated, just transform the stream.

What do you think about this?

acdlite:
This would also enable the use of handlers that accept multiple arguments.

jeron-diovis:
I'm not sure that mapping is the most often usecase. Think about validation, for example: it's very likely you'll want at least to filter your stream. Or distinct values for autocomplete requests. Or god knows what else. The core idea is to get from `createEventHandler` a ready-to-use stream, already implementing whatever logic I need here and know. And restricting this ability, forcing developer to write for same purposes a different code, as for me, would be annoying. It's like trying to be "too smart", which contradicts to `recompose`'s straightforward API.

jeron-diovis:
> This would also enable the use of handlers that accept multiple arguments.

Not sure I understand what do you mean, can you give an example?

acdlite:
The reason I say mapping would be the most common use case is because form handlers usually need to get the value of the input from the event:

```js
const { stream: value$, handler: onChange } = createEventHandler(event => event.target.value)
```

> It's like trying to be "too smart", which contradicts to recompose's straightforward API.

I don't get why transforming the stream inside `createEventHandler()` is any more straightforward than doing it outside. Could you elaborate?

> Not sure I understand what do you mean, can you give an example?

You could use the initializer function to combine multiple arguments into a single value to be pushed onto the stream:

```js
const { stream, handler }  = createEventHandler((a, b, c) => a + b + c)
stream.subscribe(val => console.log(val)
handler(1, 2, 3) // logs `6`
```

Not sure how often this would actually be useful, but it is something that's not currently possible.

jeron-diovis:
> form handlers usually need to get the value of the input from the event

Yes, of course I understand this. But then you may need to transform it, and I don't think that such need is rare enough. 

> why transforming the stream inside createEventHandler() is any more straightforward than doing it outside

Because it will force us to introduce additional variables for transformed streams:
```js
const { handler, stream } = createEventHandler(someMapper)
const transformedStream = stream.filter(...)
doStuff1(transformedStream)
doStuff2(transformedStream.someAnotherTransform())
...
```
Cases like this are pretty often. And `transformedStream` here has absolutely no sense except of being an intermediate variable. Why do we need to create at all, when it's possible to make a `stream` to already contain exactly we need?

> You could use the initializer function to combine multiple arguments into a single value to be pushed onto the stream
> it is something that's not currently possible

Why it's not possible? In your example, just pass an array to `handler`, and adopt mapper respectively.
And this will be ideologically correct as for using reactive streams. It's an event handler, it's designed to accept an event, a single object. Don't bloat up an API like this, please.

In general: you propose to overload initializer with some narrowly focused sugar, I propose to simply use it as a general-purpose extension point. I'm voting for simplicity, although, of course, it's up to you to make a final decision.

acdlite:
> Because it will force us to introduce additional variables for transformed stream

You'd just need one extra variable.

> In your example, just pass an array to handler, and adopt mapper respectively.

That requires changing the signature of the handler. I get this is an edge case, but still, it's something that's not possible with the current API.

> In general: you propose to overload initializer with some narrowly focused sugar, I propose to simply use it as a general-purpose extension point.

If simplicity is the goal, I'd prefer to leave the API as-is. To me, this:

```js
const { handler, stream } = createEventHandler(transform)
```

is less straightforward and "simple" than this:

```js
const { handler, stream } = createEventHandler(transform)
const stream2 = transform(stream)
```

It's providing two ways to do the exact same thing.

jeron-diovis:
The goal  is to get from `createEventHandler` a ready-to-use stream. Current API does not provide such ability.

jeron-diovis:
> You'd just need one extra variable

If I was not worried about polluting code with that extra variables again and again, I would not to create this PR =)

jeron-diovis:
> That requires changing the signature of the handler.

It's always possible to use helper like [`spread`](https://lodash.com/docs#spread).

acdlite:
I don't think you and I are going to agree :D We're just repeating our initial positions again.

I'm still not convinced that avoiding a single extra variable is worth expanding the scope of the API. Thanks for the PR, but I'm going to close it for now.

PR #172, opened 2016-05-18
Update to RxJS 5

### LINKED ISSUE #171: RxJS 5 support for rx-recompose
Is `rx-recompose` going to get support for RxJS 5 anytime soon? I would like to use `rx-recompose` but am unsure if it interops cleanly.

calebmer:
Closes #171.

### DIFF
--- a/src/packages/rx-recompose/createComponent.js
+++ b/src/packages/rx-recompose/createComponent.js
@@ -1,5 +1,5 @@
 import { Component } from 'react'
-import { Observable } from 'rx'
+import { Observable } from 'rxjs'
 import { createChangeEmitter } from 'change-emitter'
 
 const createComponent = propsToVdom =>
@@ -10,7 +10,7 @@ const createComponent = propsToVdom =>
 
     // Stream of props
     props$ = Observable.create(observer =>
-      this.propsEmitter.listen(props => observer.onNext(props))
+      this.propsEmitter.listen(props => observer.next(props))
     );
 
     // Stream of vdom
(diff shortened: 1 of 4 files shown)


src/packages/rx-recompose/createComponent.js
```diff
@@ -1,5 +1,5 @@
 import { Component } from 'react'
-import { Observable } from 'rx'
+import { Observable } from 'rxjs'
```
wuct: We can import only Observable

```JavaScript
import { Observable } from 'rxjs/Observable'
```

for bundle size optimization

acdlite:
Thanks for the PR but I want to wait until RxJS 5 is out of beta. At that time, we'll drop the dependency on RxJS and make it work with any spec-compliant observable. I'm also considering retiring rx-recompose entirely and bringing it into the main project.

See the rx-recompose-next branch for peek at how this will look. A bit out of date, but some version of that is what will eventually be released.

calebmer:
That's a good goal. While RxJS isn't out of beta, the spec *is* finalized. Whether this means creating a new `recompose-observable` or rolling it into `recompose` can be done now for the people using the spec. RxJS 5 is in a pretty stable condition at the moment.

I'd like to see the plan to update this module implemented soon for early adopters. What can I do to help?

istarkov:
I also see that RxJs 5 is more widely used in that projects I'm interested in, so also think that even it is in beta it will be the best choice to use it now.

chicoxyzzy:
fyi while rxjs 5 api is frozen the library is not stable enough yet

acdlite:
@calebmer I went ahead and merged your branch into `rx-recompose-next`. Added an additional commit to remove the dependency on RxJS: https://github.com/acdlite/recompose/commit/5214295d12f1c3ef23e98a734a51cdba6e5ffa2e

Care to review it? Then I'll publish it on npm with a `next` tag.

@chicoxyzzy Thanks for volunteering! Care to review as well?

chicoxyzzy:
yep. I'll try this branch but not earlier than week after ReactEurope

PR #173, opened 2016-05-18
fix(docs): fix mapProps callback in addCounting example

justingreenberg:


wuct:
Thanks @justingreenberg

PR #174, opened 2016-05-18
Adding addPropTypes method to combine proptypes.

psachs21:
Creates a higher level component that adds proptypes to existing props. Useful for using in other high level components that need to maintain existing proptypes.


acdlite:
> Useful for using in other high level components that need to maintain existing proptypes.

Why not set the propTypes on the HoC's component, rather than add them to the base component?

psachs21:
This may be because I'm still pretty new to recompose and there may be a better way of doing this, but I have one HoC that needs the prop types, and I'd like to compose it with another that needs it's own prop types. If i understand correctly, setting prop types wipes out any previous proptypes.

Here's an example of what I mean:
```js
//utils.js
export const withContainerQuery = (query, key) => {
  return compose(
    addPropTypes({
      setContainerQuery: React.PropTypes.func.isRequired,
      clearContainerQuery: React.PropTypes.func.isRequired
    })
  );
};
...
// messageList.js
const enhance = compose(
  addPropTypes({
    messageList: PropTypes.any.isRequired
  }),
  withContainerQuery(query, 'messageList')
);
```
I'd obviously like to be able to compose withContainerQuery without having to know the details about the propTypes in requires.

wuct:
IMHO it would be more clear if you export `withContainerQuery`'s `propTypes` and import it explicitly 

```JavaScript
//utils.js
export const propTypes = {
  setContainerQuery: React.PropTypes.func.isRequired,
  clearContainerQuery: React.PropTypes.func.isRequired
};

export const withContainerQuery = (query, key) => {
  return compose(
    setPropTypes(propTypes)
  );
};
...
// messageList.js
import { propTypes: withContainerQueryPropTypes } from 'withContainerQuery'

const enhance = compose(
  setPropTypes({
    ...withContainerQueryPropTypes,
    messageList: PropTypes.any.isRequired
  }),
  withContainerQuery(query, 'messageList')
);
```

SeeThruHead:
Hey wuct, I just wanted to chime in and add a little context.

We're using react-redux-provide here, so propTypes are actually used to signal what information we need from the store. We only use those proptypes inside the HOC and as such they are just an implementation detail of the HOC. (We would never manually pass those proptypes) If we use addPropTypes everywhere it makes it so that we never have to know what pieces of data our HOC take from the store to do what they have to do. We can safely compose HOC without knowing how they work internally.

I would agree with setPropTypes if these were props the developer had to pass when rendering the component, as that documents the api of the component (though i think we can be sure that a component composed of multiple HOC would have an api that is the composition of those HOC), but we don't in this instance, and want to save the developer from having to worry about the internal workings of our HOC.

wuct:
Thanks for elaborating. Now I can get your point.

Considering naming consistency, `withPropTypes` might be a better name for this HOC.

Besides, maybe we can modify current `setPropTypes`'s signature to

```JavaScript
setPropTypes(
  propTypes: Object | (baseComponentPropTypes: Object) => Object
): HigherOrderComponent
```

which is like `withState`'s [`initailState`](https://github.com/acdlite/recompose/blob/master/docs/API.md#withstate).

Then we can use it like

```JavaScript
setPropTypes(baseComponentPropTypes => ({
  ...baseComponentPropTypes,
  // set other propTypes here
}))
```

How do you think?

SeeThruHead:
I like the naming change.
Would you suggest changing setPropTypes to handle both and then then creating a withPropTypes that uses it for consistent naming?
Or change setPropTypes to withPropTypes? (Probably not great since once use of setPropTypes wouldn't follow the 'with' convention)

psachs21:
I think your first option makes the most sense. setPropTypes can also take a function and withPropTypes  will just call setPropTypes.

psachs21:
Although, looking at the source, it does look silly to have withPropTypes be just an alias.

wuct:
IMHO `withPropTypes` can be a shortcut for `setPropTypes`

```JavaScript
const withPropTypes = propTypes =>
  setPropTypes(baseComponentPropTypes => ({
    ...baseComponentPropTypes,
    ...propTypes
  }))
```

As a result, the behaviors of `setPropTypes` and `withPropTypes` would be different.

```JavaScript
setPropTypes(foo)
```
replaces the base component propTypes with `foo`.

And, 

```JavaScript
withPropTypes(foo)
```

just appends `foo` to the base component propTypes.

PR #179, opened 2016-05-20
Update eslint

acdlite:


istarkov:
Our last eslint + airnbnb config update cost us 8000+ code lines change ;-) For some files history becomes useless ;-)

PR #181, opened 2016-05-21
Small tidy up `withHandlers` usage example

justingreenberg:
- removed `<label>` since not necessary
- removed 'value' prop from `<input />` since initial state is empty 
- render value to demo state

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -151,11 +151,10 @@ const enhance = compose(
   })
 )
 
-const Form = enhance(({ value, onChange, onSubmit }) =>
+const Form = enhance(({ onChange, onSubmit, value }) =>
   <form onSubmit={onSubmit}>
-    <label>Value
-      <input type="text" value={value} onChange={onChange} />
-    </label>
+    <input type="text" onChange={onChange} />
+    <p>Value: {value}</p>
   </form>
 )
 ```


istarkov:
Hi, Thank you for contributing.

`value` must be present in `input` as it is the only combination to make `input` controllable,
please read about controlled components here: http://facebook.github.io/react/docs/forms.html#controlled-components

justingreenberg:
thank you—however, the `withState` HOC is providing control of the component... 

please see:
https://github.com/acdlite/recompose/blob/master/src/packages/recompose/withState.js#L9 
https://jsfiddle.net/whiskypeters/s7y1bko4/

justingreenberg:
edit: i'm sorry for the rushed explanation this morning! we don't need to control the form components since state is delegated to `withState` HOC, which exposes the state and updater function as props... by composing with `withHandlers` HOC, we can process events and update state via props closure

we may want to pass a value to `<input>` to bind the initial state, however in that case we should be using `defaultValue` as the component is uncontrolled

this conversation demonstrates how the current example can cause confusion!

istarkov:
There are no confusion. It is just a simple usage example, without edge cases. I don't think that it should be changed as in current PR. I always use all controls as controlled, so don't think that example with uncontrollable input is a good idea. Initial state for value is set in withState, so I can't get your problem.

justingreenberg:
doing some housecleaning, closing this :)

Issue #97, opened 2016-02-11
Smart compose


istarkov:
Some developers I work with dislike that `recompose` creates HOCs on each operation, so the tree in react tools is too deep.

As an example:
This code
https://github.com/istarkov/google-map-clustering-example/blob/master/src/GMap.js#L57-L126

creates this tree
![image](https://cloud.githubusercontent.com/assets/5077042/12974970/4c6b7d3a-d0c9-11e5-9b92-9cee3b015f8c.png)

The real examples we have at work has more deeper trees. (I does not use react tools at work, but others use it a lot)

And Yes I know that I can combine here 3 call to `withState` into one or/and `mapProps..` but this affects code readability for me.

I think that some of that multiple calls can easily (or not ;-)) be combined at compose level, as some optimization operation. 
For example sequential calls to `mapPropsOnChange` can be combined into one HOC automatically and etc....

May be it's too early to talk about this, but it's just an opinion of some developers.

acdlite:
We address this in part using our custom version of [`createElement()`](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/createElement.js), which avoids unnecessary intermediate React elements by detecting referentially transparent function components. This only works with certain helpers like `mapProps()` — helpers that use lifecycle methods, context, etc necessarily require a class component.

So there's only so much we can do here. Better documentation and examples that illustrate best practices is needed.

For example, instead of using two sequential `mapPropsOnChange()`, use a single `mapPropsOnChange()` and compose the mapping functions together.

Instead of multiple `withState()`, use a single `withState()` or use `withReducer()`.

Or if your component is sufficiently complex and the extra elements concern you, then perhaps Recompose isn't the right fit for that particular component and you should use a regular class instead.

acdlite:
See https://github.com/acdlite/recompose/issues/182 for a potential solution to this problem

PR #182, opened 2016-05-21
RFC: Update middleware

acdlite:
As `mapPropsStream` shows, many (most?) higher-order components can be described in terms of a props stream transformation. If multiple such higher-order components are composed together, we reduce the number of components by composing the transformations together instead, which should be more lightweight than sending props through React's update system at each level. In other words, the following two forms are equivalent:

``` js
// Wraps the base component in three container components
compose(
  mapPropsStream(transformA),
  mapPropsStream(transformB),
  mapPropsStream(transformC)
)(BaseComponent)

// Same as this, which wraps the base component in a single
// container component
mapPropsStream(
  compose(transformA, transformB, transformC)
)(BaseComponent)
```

But we don't want to require that Recompose users include an observable pollyfill. Instead, I'm proposing a lightweight abstraction called "update middleware" which will serve the same purpose.

So far, I've implemented `withState` and `withHandlers` using this pattern.

If we go in this direction, we will modify the compose function to detect HOCs defined using middleware and compose them using the second strategy from the example above.

### DIFF
--- a/src/packages/recompose/flattenProp.js
+++ b/src/packages/recompose/flattenProp.js
@@ -1,15 +1,11 @@
+import mapProps from './mapProps'
 import omit from './utils/omit'
 import createHelper from './createHelper'
-import createEagerFactory from './createEagerFactory'
 
-const flattenProp = propName => BaseComponent => {
-  const factory = createEagerFactory(BaseComponent)
-  return props => (
-    factory({
-      ...omit(props, [propName]),
-      ...props[propName]
-    })
-  )
-}
+const flattenProp = propName =>
+  mapProps(props => ({
+    ...omit(props, [propName]),
+    ...props[propName]
+  }))
 
 export default createHelper(flattenProp, 'flattenProp')
--- a/src/packages/recompose/utils/createHocFromMiddleware.js
+++ b/src/packages/recompose/utils/createHocFromMiddleware.js
@@ -80,7 +80,6 @@ const createComponentFromMiddleware = (middlewares, BaseComponent) => {
     middlewares,
     BaseComponent
   }
-  CreateComponentFromMiddleware.middlewares = middlewares
   return CreateComponentFromMiddleware
 }
 
--- a/src/packages/recompose/utils/createHocFromMiddleware.js
+++ b/src/packages/recompose/utils/createHocFromMiddleware.js
@@ -27,7 +27,7 @@ const createHocFromMiddleware = (...middlewares) => {
             }
           })
           return xforms
-        }, {})
+        }, [])
       }
 
       state = { childProps: null }
--- a/src/packages/recompose/utils/createHocFromMiddleware.js
+++ b/src/packages/recompose/utils/createHocFromMiddleware.js
@@ -1,7 +1,7 @@
 import { Component } from 'react'
-import createEagerFactory from './createEagerFactory'
+import createEagerFactory from '../createEagerFactory'
 
-const applyUpdateMiddleware = (...middlewares) => {
+const createHocFromMiddleware = (...middlewares) => {
   const higherOrderComponent = BaseComponent => {
     const factory = createEagerFactory(BaseComponent)
     return class extends Component {
@@ -80,4 +80,4 @@ const applyUpdateMiddleware = (...middlewares) => {
   return higherOrderComponent
 }
 
-export default applyUpdateMiddleware
+export default createHocFromMiddleware
--- a/src/packages/recompose/utils/createHocFromMiddleware.js
+++ b/src/packages/recompose/utils/createHocFromMiddleware.js
@@ -7,7 +7,6 @@ const createHocFromMiddleware = (...middlewares) => {
     return class extends Component {
       constructor(initialProps) {
         super(initialProps)
-        this.xforms = []
         this.pendingXformProps = [initialProps]
         this.xformProps = []
 
--- a/src/packages/recompose/utils/createHocFromMiddleware.js
+++ b/src/packages/recompose/utils/createHocFromMiddleware.js
@@ -1,82 +1,97 @@
 import { Component } from 'react'
 import createEagerFactory from '../createEagerFactory'
 
-const createHocFromMiddleware = (...middlewares) => {
-  const higherOrderComponent = BaseComponent => {
-    const factory = createEagerFactory(BaseComponent)
-    return class extends Component {
-      constructor(initialProps) {
-        super(initialProps)
-        this.pendingXformProps = [initialProps]
-        this.xformProps = []
+const MIDDLEWARE_INFO = typeof Symbol === 'function'
+  ? Symbol()
+  : '@@recompose/middlewareInfo'
 
-        this.xforms = middlewares.reduce((xforms, middleware, i) => {
-          const createXform = middleware({
-            getProps: () => this.xformProps[i]
-          })
-          xforms[i] = createXform({
-            update: (props, cb) => {
-              this.xformProps[i] = this.pendingXformProps[i]
-              this.pendingXformProps[i + 1] = props
-              if (i === middlewares.length - 1) {
-                this.applyPropUpdate(props, cb)
-              } else {
-                this.xforms[i + 1].update(props, cb)
-              }
+const createComponentFromMiddleware = (middlewares, BaseComponent) => {
+  const factory = createEagerFactory(BaseComponent)
+  class CreateComponentFromMiddleware extends Component {
+    constructor(initialProps) {
+      super(initialProps)
+      this.pendingXformProps = [initialProps]
+      this.xformProps = []
+
+      this.xforms = middlewares.reduce((xforms, middleware, i) => {
+        const createXform = middleware({
+          getProps: () => this.xformProps[i]
+        })
+        xforms[i] = createXform({
+          update: (props, cb) => {
+            this.xformProps[i] = this.pendingXformProps[i]
+            this.pendingXformProps[i + 1] = props
+            if (i === middlewares.length - 1) {
+              this.applyPropUpdate(props, cb)
+            } else {
+              this.xforms[i + 1].update(props, cb)
             }
-          })
-          return xforms
-        }, [])
-      }
-
-      state = { childProps: null }
-      _isMounted = false
-      didReceivePropUpdate = false
-
-      destroy() {
-        this.xforms.forEach(xform => {
-          if (typeof xform.destroy === 'function') {
-            xform.destroy()
           }
         })
-      }
+        return xforms
+      }, [])
+    }
 
-      applyPropUpdate(props, cb) {
-        if (!this._isMounted) {
-          return
+    state = { childProps: null }
+    _isMounted = false
+    didReceivePropUpdate = false
+
+    destroy() {
+      this.xforms.forEach(xform => {
+        if (typeof xform.destroy === 'function') {
+          xform.destroy()
         }
-        this.didReceivePropUpdate = true
-        this.setState({
-          childProps: props
-        }, cb)
-      }
+      })
+    }
 
-      componentWillMount() {
-        this._isMounted = true
-        this.xforms[0].update(this.pendingXformProps[0])
+    applyPropUpdate(props, cb) {
+      if (!this._isMounted) {
+        return
       }
+      this.didReceivePropUpdate = true
+      this.setState({
+        childProps: props
+      }, cb)
+    }
 
-      componentWillReceiveProps(nextProps) {
-        this.pendingXformProps[0] = nextProps
-        this.xforms[0].update(nextProps)
-      }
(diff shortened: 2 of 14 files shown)


acdlite:
@wuct @istarkov

src/packages/recompose/flattenProp.js
```diff
@@ -1,15 +1,11 @@
...
-      ...omit(props, [propName]),
-      ...props[propName]
-    })
-  )
-}
+const flattenProp = propName =>
+  mapProps(props => ({
+    ...omit(props, [propName]),
+    ...props[propName]
+  }))
```
acdlite: In case it's not clear why this is included in this PR (other than being the right way to define `flattenProp`, anyway), the idea is that there are some helpers that we don't want to define using middleware, because middleware necessitates using a class component, even when a functional component would suffice. What we really want is to use a functional component *unless* the HOC is composed with another HOC which *must* use a class. Put another way, a single functional component is better than a class component, but a single class component is better than a class component and a functional component.

To achieve this, we will special case `mapProps` to use a functional component by default, but use middleware when being composed with other middleware-using HOCs.

acdlite:
Also, to be clear, this abstraction would be private: it would not be exposed to the public, except to advertise as a performance optimization.

src/packages/recompose/withPropsOnChange.js
```diff
@@ -1,34 +1,37 @@
...
-    computedProps = propsMapper(this.props);
+  return createHocFromMiddleware(({ getProps }) => next => {
+    let didUpdate = false
+    let computedProps = null
 
-    componentWillReceiveProps(nextProps) {
-      if (shouldMap(this.props, nextProps)) {
-        this.computedProps = propsMapper(nextProps)
-      }
+    const update = (props, cb) => {
```
wuct: No need to create a `update` function here.
We can move `next.update(...)` to line 31.

src/packages/recompose/withHandlers.js
```diff
@@ -44,19 +42,17 @@ const withHandlers = handlers => BaseComponent => {
...
-
-    render() {
-      return factory({
-        ...this.props,
-        ...this.handlers
-      })
+    )
+
+    return {
+      update: (props, cb) => {
```
wuct: Considering naming consistency, `nextProps` might be better than `props`

wuct:
I really this new middleware feature. I believe the new `compose` function could also solve #97 😄

acdlite:
Yeah, exactly. The whole point is to solve #97.

acdlite:
This is working out better than I thought Turns out we don't even need a smart compose: https://github.com/acdlite/recompose/blob/132fe2f25713a5ee5fd16366afeaf75576e96b30/src/packages/recompose/utils/createHocFromMiddleware.js#L88-L93

src/packages/recompose/utils/createHocFromMiddleware.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    state = { childProps: null }
+    _isMounted = false
+    didReceivePropUpdate = false
+
+    destroy() {
+      this.xforms.forEach(xform => {
+        if (typeof xform.destroy === 'function') {
+          xform.destroy()
```
istarkov: So the only need the middleware returns object is `.destroy`?

acdlite: Middleware returns an object (right now I'm calling it `xform` for "transform" but that's not necessarily final) with a (required) `update` function. It may also have a `destroy` function to clean up subscriptions, though this isn't being used yet by any of the HOCs I've implemented so far. `mapPropsStream` is an example of an HOC that would need it.

istarkov:
Just investigated a little, and looks like rxjs 5 does not give a lot overhead on dist size, 
[here](https://github.com/istarkov/recompose-2-rx/blob/master/src/index.js) is an example of `mapProps, withState, withHandlers` based on `rxjs`, the result distributive size is 28kb minified, 7kb zipped.

And yes all this functions is written over `props$` stream so should be used like

```javascript
mapPropsStream(flow(
     mapProps(...), 
     withState(...) 
))(BaseComponent)
```

[example](https://github.com/istarkov/recompose-2-rx/blob/master/tests/index.spec.js)

acdlite:
But what about the cost of importing a single module? I want to keep that as small as reasonably possible so that library authors can rely on individual helpers guilt free.

acdlite:
Just an update: this PR is on hold pending further investigation. It's not clear to me there's any performance benefit to squashing components this way.

If/when it is clear that there is a significant performance boost, we will reevaluate.

wuct:
We should open an issue for adding performance testing

rtsao:
There's still a benefit in terms of usability so even if there was no performance benefit I think it would still be worthwhile.
- Devtools ballooning
- Needing traverse through all wrappers when accessing refs or instance methods of a component

neoziro:
This is the approach used in [recompact](https://github.com/neoziro/recompact). I did some benchmarks, you can find it in the repo. Some conclusions from benchmarks: 

- At mount it is better only if you stack a lot of hoc
- At update it's always better, event with a small number of hoc
- Performance are never degraded in comparison of recompose

However there is a true benefit in devtools.

jeffbski:
@neoziro Thanks for mentioning your project. It would be nice if you have a way to view the  benchmark results without having to install and run them from your repo. Also include the summary like you have provided here.

acdlite:
Note: this is also on hold until the incremental rendering stuff we're doing in Fiber shakes out a bit more. (Goes for all of Recompose, not just this PR.) Going to have to revaluate parts of the API. Apologies :)

acdlite:
To add a bit more context, the observable utils as currently written won't work with Fiber because they assume that every `componentWillReceiveProps` corresponds to a flushed render. In Fiber, renders can be aborted before they are flushed. We'll need to think of a new API for Recompose that can properly handle side-effects.

neoziro:
OK, it's not exactly the same implementation in recompact, but we probably have the same problems. Let's wait and try when Fiber will be ready!

istarkov:
> To add a bit more context, the observable utils as currently written won't work with Fiber

@acdlite Can you explain a bit of more (_may be with small example_) what issues it can cause, and how to solve them (In my understanding the similar issues will occur for all components where internal state depends on props)

cvle:
@acdlite I'd love to have more information too. 

I use _recompose_ extensively in my projects and I really love it. Though the problems with Dev Tools Ballooning and performance degradation in complex compositions especially on lower end mobile devices was becoming a huge issue.

So I started with a proof of concept to fix dev tools ballooning and to speed up my _recompose_ which is now [reassemble](https://github.com/wikiwi/reassemble). It just uses a collection of callbacks called _Composables_ instead of HOCs that are merged into a single HOC. You might be interested in a performance comparison of the three projects: [recompose, recompact and reassemble](https://github.com/wikiwi/reassemble/tree/master/benchmark)

istarkov:
@cvle Do you run the tests in production mode? With NODE_ENV=production?

cvle:
@istarkov Good catch! I forgot to do that.

cvle:
Updated Benchmarks with interesting results.

neoziro:
@cvle can you update it with recompact benchs? I think we should put effort together and merge our projects, what do you think?

neoziro:
The approach of reassemble is different, you are not exposing any hoc and it's not a drop-in replacement for recompose even if the API seems the same.

Also you replace `compose` that is like a native approach is composition, users should be able to write:

```js
pure(withProps({ foo: 'bar' })(Component))
// or 
compose(
  pure,
  withProps({ foo: 'bar' }),
)(Component)
```

You lose the composition aspect of hoc. It reminds me of [mixout](https://github.com/alitaheri/react-mixout).

Also I updated [recompact benchmarks](https://github.com/neoziro/recompact/tree/master/src/__benchmarks__) and published the results (removing enzyme was the solution to stabilize them). [My results](https://github.com/neoziro/recompact/tree/master/src/__benchmarks__#results) are very different from yours:

cvle:
@neoziro As I mentioned _reassemble_ is not using HOCs. I changed the project description to be more clear. I'm still puzzled by the differences in our two benchmarks. I can't spot the cause, anyone?

I like the idea of _recompact_, in fact I considered a similar implementation. Atm I'm using _reassemble_ in some of my projects to speed up initial render. Though I need to do more performance testing. I'll keep an eye on _recompact_.

cvle:
@neoziro Your benchmark is incorrect. You need to cleanup on each run, otherwise the React reconciler kicks in prevent any rerendering.

cvle:
Another difference between _recompact_ and _reassemble_ is that _recompact_ cannot squash context related HOCs and the Lifecycle HOC. I often use lifecycle hooks for deferred rendering and to deal with animations (e.g. in my other open source project: [react-css-transition](https://github.com/wikiwi/react-css-transition)).

neoziro:
@cvle you are right, I will update benchmarks soon.

Fried-Chicken:
@acdlite Will this be PR be merged in soon?

istarkov:
@Fried-Chicken there are no reason for this now. The only issue is a React dev tools tree length.
There are no big performance gain as vdom works really fast when you have a pack of components with only one child. May be some real life (_not a tests like lets compare 10 plain `withState` perf vs 10 recompose `withState`_) exists but we have not seen them.
If you really need (really?) this you can use https://github.com/neoziro/recompact as drop in replacement for recompose or similar solutions.

istarkov:
@Fried-Chicken benchmarks https://github.com/neoziro/recompact/tree/master/src/__benchmarks__

atomanyih:
@istarkov 
If the work has already been done, it doesn't make things slower, and it makes it easier for people to develop with recompose, why not include it?

I love recompose and use it frequently, but it can make debugging devtools kind of incomprehensible. This is especially true for other people on a team who didn't write the recomposed components and have to crawl through the component tree to try to figure out which file contains one particular occurrence of `withProps(withProps(withProps(UiComponent)))`

istarkov:
I dont use devtools so this is not a problem for me. And I love recompose because of simplicity of its code, IMO its so simple that does not need a docs. This change will make code less readable. So for me there are no advantages to apply this.

smeijer:
> I dont use devtools so this is not a problem for me.... So for me there are no advantages to apply this.

Sorry, but as a maintainer of a project with 7k stars on github, and ~ 780.000 downloads each month from npm, this response feels kind of wrong. 

Don't get me wrong. `recompose` is an awesome library, and the amount of stars and downloads support my statement. But the ballooning is a problem.

The only reason for me to not switch over to `recompact` or `reassemble`; is that this project seems to be more alive / maintained.

istarkov:
How amount of stars or downloads can affect decisions? It's impossible to make project good for all so a metric "good for me" is very very good. If we'll start to accept every PR and every suggestion here it will be unusable project.

rbong:
The issue of devtools has been brought up, but not the one of serialization, particularly for snapshot tests with enzyme and Jest.

When the component tree is so polluted, snapshot tests are going to change often, and the change diffs are not going to be readable. This makes snapshot testing too volatile to use in many cases. I know that this seems like a pretty specific use case, but [500,000+ people have downloaded enzyme-to-json in the past month](https://www.npmjs.com/package/enzyme-to-json).

Are there no workarounds, or am I wrong?

istarkov:
@rbong have you checked your words? Have you seen the snapshot generated by enzyme 2 json? 
If yes please answer what are you see in snapshot?

neoziro:
@rbong I think it is better to let recompose a simple toolkit of hoc utilities (as simple as possible). If you want to use it as a framework and stack several hocs you should take a look to [recompact](https://github.com/neoziro/recompact). I am not very active on it lately, but it is still maintained and I am open to PR.

istarkov:
@neoziro the @rbong problem is not exist, its obvious if you just look into the snapshot. As snapshot is not the vdom tree, and independent of amount of hocs used.

neoziro:
@istarkov I mean for dev tools or other issues. You should lock this conversation.

deepsweet:
I think it's a matter of creating a special 3rd party Babel plugin to convert `compose()` and base HOCs to something else (if it's possible at all ofc).

smeijer:
I think such a plugin would be awesome. But I'm not sure if it would work. This because `recompact` and `reassembly` work differently when you need something like `redux`.

In this sample, `dispatch` is `undefined` in both `recompact` and `reassemble`, but it does work in `recompose`.

```js
export default compose(
  connect(),
  withHanders({
    onClick: ({ dispatch }) => () => {
      dispatch('foo');
    },
  }),
)
```

rbong:
@istarkov my apologies, I did not know enzyme-to-json contained a function to only render DOM nodes.

istarkov:
@rbong not a problem :-) no need in apologies

PR #184, opened 2016-05-25
Test immutability of handlers passed by withHandlers

matthieuprat:
It seems like the immutability of handlers passed by `withHandlers` is not tested, is it?


wuct:
Thanks! Looks good to me

Issue #168, opened 2016-05-17
Support for throttleProp and debounceProp


wuct:
## throttleProp

```
throttleProp(
  propName: string,
  wait: number 
): HigherOrderComponent
```

Only update a prop at most once per every `wait` milliseconds. 

`debounceProp` is similar. Might be useful when a prop is updated frequently. How do you think?

acdlite:
Throttling and debouncing are good ideas, but what do you mean by throttling a specific prop? Either the whole component updates or not, right?

acdlite:
Oh I think I see what you mean:

- When the component receives an update, it checks which props have changed using strict equality comparison.
- If any non-throttled props have changed, those props updates are applied immediately.
- If the throttled prop has changed, it is applied using a throttled update method.

Is that right?

wuct:
@acdlite Yes. In my mind it will be something like this

```
 mapPropsStream(props$ =>
  Observable.combineLatest(
    props$
      .pluck('throttledProps')
      .distinctUntilChanged()
      .throttle(900),
    props$
      .map(({ throttledProps, ...otherProps }) => otherProps)
      .distinctUntilChanged(undefined, shallowEqual),
    (throttledProps, otherProps) => ({ ...otherProps, throttledProps })
  )
)
```

but without Observable.

wuct:
I can implement them if they look good

acdlite:
Sure!

acdlite:
@wuct Maybe implement these with middleware and submit the PR to my `update-middleware` branch?

wuct:
@acdlite sure. I will find some time for it this week :)

deepsweet:
Hey guys, just implemented these two as a part of my HOCs collection – https://github.com/deepsweet/hocs

gastonmorixe:
With lodash

```js

const throttleHandler = (
  propName,
  delay,
  options = {
    leading: true,
    trailing: true
  }
) => WrappedComponent =>
  class extends React.PureComponent {
    original = () => {
      this.props[propName]()
    }

    originalThrottled = throttle(this.original, delay, options)

    replace = {
      [propName]: this.originalThrottled
    }

    render() {
      return <WrappedComponent {...this.props} {...this.replace} />
    }
  }
```

PR #185, opened 2016-05-29
Support for throttleProp

wuct:
Please see #168 for more detail


src/packages/recompose/__tests__/throttleProp-test.js
```diff
@@ -0,0 +1,68 @@
...
+test.cb('throttleProp cancels the latest update before destroyed', t => {
+  t.plan(1)
+
+  const Counter = throttleProp('throttled', 30)('div')
+  const wrapper = mount(<Counter throttled={0} />)
+  const div = wrapper.find('div')
+
+  wrapper.setProps({ throttled: 1 })
+
+  wrapper.instance().xforms[0].destroy()
```
wuct: I do not think this is a good idea to introduce `xforms` here, because `xforms` is designed for internal use in `createHocFromMiddleware` and might be changed in the future. 
However, it is just a quick hack and advices are welcome.

acdlite:
Cool! (Heads up, I probably won't be able to review this until next week when I get back from React Europe.)

wuct:
@acdlite Good luck on your talk and workshop ;)
I wish I could be there.

istarkov:
@wuct having `mapPropsStream` do we need `throttleProps`?

wuct:
@istarkov IMHO it still need time for Observable to get widely-adopted, so `throttleProps` might be still useful.

However, `mapPropsStream` can replace all other functions in the end.

jschlieber:
When can we expect this feature to be released?

wuct:
@jschlieber Because this PR is blocked by #182, the time is still uncertain.

istarkov:
Will close as corresponding or is also closed, and it looks to me that its always better to throttle handlers and not props as like as here https://github.com/deepsweet/hocs

PR #189, opened 2016-06-03
loosen peer dependency restriction on relay

brettjashford:
would like to use this util w/ relay 0.8.0


istarkov:
Thank you! 
Will close as a duplicate of #234

Issue #190, opened 2016-06-05
Flatten Prop without ommiting prop


Secretmapper:
Using `flattenProp` flattens the property, and recompose (intuitively) removes the flattened property.

I find however, that more often than not I need to access the base, unflattened prop while also appreciating the convenience of a flattened prop.

As an example, consider a Relay container like so:
```
 fragment on Item {
  fieldOne,
  fieldTwo
  ${SubItem.getFragment('item')}
}
```

and a component:
```javascript
<div>
  <div>{fieldOne}</div>
  <div>{fieldTwo}</div>
  <SubItem item={item} />
</div>
```

Using flattenProp prevents me from doing the above, and I need to instead opt for the uglier:
```javascript
<div>
  <div>{item.fieldOne}</div>
  <div>{item.fieldTwo}</div>
  <SubItem item={item} />
</div>
```

It would be great if it is possible to flattenProp without omitting the actual property. It can even be an optional argument to `flattenProp` to maintain backwards compatibility, or even a new HOC if deemed necessary.

I'm not sure if there are complications that might arise from this, or if there are already existing ways to do this with the current API.

istarkov:
With modern js it's not a big deal to write flattenProp you need using `mapProps`.

```javascript
mapProps(({ ...props, item }) => ({ ...props, ...item, item }))
```

acdlite:
This is a good point. I can't imagine having the original prop stick around is ever a problem.

acdlite:
Closed by #195.

PR #195, opened 2016-06-10
changed flattenProp hoc it does not omit flattened prop

yjcxy12:
Connects #190 

There is one issue that has not resolved. If the flatten prop has a key with the same name of the prop, it will be overwritten. As shown on second test case of `flattenProp-test.js`.

Any suggestion on solving this?

### DIFF
--- a/src/packages/recompose/__tests__/flattenProp-test.js
+++ b/src/packages/recompose/__tests__/flattenProp-test.js
@@ -7,11 +7,18 @@ test('flattenProps flattens an object prop and spreads it into the top-level pro
   const Counter = flattenProp('state')('div')
   t.is(Counter.displayName, 'flattenProp(div)')
 
-  const wrapper = shallow(
+  let wrapper = shallow(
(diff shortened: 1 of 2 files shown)


acdlite:
In most cases I can't imagine this will be a problem. In those cases where it is, we should direct people to use `mapProps`.

src/packages/recompose/__tests__/flattenProp-test.js
```diff
@@ -7,11 +7,18 @@ test('flattenProps flattens an object prop and spreads it into the top-level pro
...
+  let wrapper = shallow(
     <Counter pass="through" state={{ counter: 1 }} />
   )
 
   t.true(wrapper.equals(
-    <div pass="through" counter={1} />
+    <div pass="through" state={{ counter: 1 }} counter={1} />
+  ))
+
+  wrapper = shallow(
```
acdlite: Use `wrapper.setProps` instead of creating a new wrapper, please

acdlite:
Thanks!

PR #196, opened 2016-06-11
Merge rx-recompose into main project

acdlite:
- [x] No RxJS dependency. Uses observables that conform to Observable spec and implement `Symbol.observable`.
- [x] Configurable interop with stream libraries like RxJS, most, xstream, and Bacon
- [x] Renames `createComponent` to `componentFromStream`


src/packages/recompose/__tests__/mapPropsStream-test.js
```diff
@@ -1,20 +1,32 @@
...
+
 const identity = t => t
 
 test('mapPropsStream maps a stream of owner props to a stream of child props', t => {
   const SmartButton = mapPropsStream(props$ => {
     const { handler: onClick, stream: increment$ } = createEventHandler()
     const count$ = increment$
-      .startWith(0)
-      .scan(total => total + 1)
+      ::startWith(0)
```
chicoxyzzy: Bind operator is still at stage 0 so syntax may change in future and there are no guarantees it will be included in spec

chicoxyzzy:
It'll be cool if we can use not only RxJS but also [most](https://github.com/cujojs/most/pull/254) and other libraries. But this means recompose should use `Symbol.observable` instead of `Observable`

chicoxyzzy:
also interesting thread 
https://github.com/cujojs/most/issues/160

src/packages/recompose/mapPropsStream.js
```diff
@@ -0,0 +1,17 @@
+import createEagerFactory from './createEagerFactory'
+import createHelper from './createHelper'
+import componentFromStream from './componentFromStream'
+
+const mapPropsStream = transform => BaseComponent => {
+  const factory = createEagerFactory(BaseComponent)
+  return componentFromStream(ownerProps$ =>
+    Observable.create(observer => {
```
istarkov: Can't see `.create` in proposal https://github.com/zenparsing/es-observable
may be better to use `new Observable(observer =>` instead

src/packages/recompose/__tests__/mapPropsStream-test.js
acdlite: Well it's a test, so oh well for now :D

acdlite:
@chicoxyzzy 

> It'll be cool if we can use not only RxJS but also most and other libraries

As long as Observable is a global that follows the spec, it should work, right?

src/packages/recompose/mapPropsStream.js
acdlite: Good catch!

src/packages/recompose/__tests__/mapPropsStream-test.js
acdlite: The reason I'm importing each helper separately is to ensure as much as possible that we're not depending on any RxJS-specific functionality.

chicoxyzzy:
Hmmm.. Probably you are right and it will work. I think I can create PR with some tests for most as well if you are not opposed =)

istarkov:
Good idea at @chicoxyzzy ref https://github.com/cujojs/most/issues/160

It will allow to remove global Observable ref, and the only downside 
before using with with `rxjs` without bind syntax is that you will need to convert 
`Rx.Observable.from(props$).hereICanUseRxJsOperatorsAsIs`

src/packages/recompose/__tests__/mapPropsStream-test.js
chicoxyzzy: yes, I understand that

chicoxyzzy:
We can still use bind syntax after converting ES-compatable stream to RxJS observable

chicoxyzzy:
Just checked out. RxJS API and [Most.js API](https://github.com/cujojs/most/blob/master/docs/api.md#mostjs-api) are not 100% compatible so current docs from PR will work only for Rx but not Most.
We really can remove global `Observable` ref but in that case we should add `[Symbol.observable]()` to Recompose streams to make an interop possible. This is good because not Rx nor Most are actually polyfills but libraries for streams (kind of lodash for streams).

chicoxyzzy:
The original `Observable` from spec just have [two methods](https://github.com/zenparsing/es-observable#api). But it's possible to easily convert then to Rx or Most like so:
```js
import Rx from 'rxjs/Rx';
import { from } from 'most';

const RxObservable = Rx.Observable.from(SomeObservable);
const MostObservable = from(SomeObservable);
```

chicoxyzzy:
Eventually this also can make integration with [mobx](https://github.com/mobxjs/mobx/issues/169) possible (for now it will be possible using `Rx.Observable.from` and https://github.com/chicoxyzzy/rx-mobx)

chicoxyzzy:
Redux does the same thing https://github.com/reactjs/redux/pull/1632

However IMO it's not so trivial to add `[Symbol.observable]()` method for Rx-Recompose and maybe it's an idea for distinct library

acdlite:
> We really can remove global Observable ref but in that case we should add [Symbol.observable]() to Recompose streams to make an interop possible. This is good because not Rx nor Most are actually polyfills but libraries for streams (kind of lodash for streams).

I'm not sure we're on the same page. What is a "Recompose stream"? Are you referring to `props$`, the stream of props that is provided to `mapPropsStream` and `componentFromStream`? My goal is avoid introducing a Recompose-specific stream type entirely by using the observable spec.

The purpose of `[Symbol.observable()]` is to turn a stream-like value into an Observable. I don't see how it applies here, given that we're using observables directly.

A code example explaining what you mean would be helpful.

acdlite:
Oh wait, I think I see what you mean about removing the global Observable. I don't think it's necessary for most compatibility, but it's a good idea nonetheless.

acdlite:
@chicoxyzzy @istarkov Alright, I just pushed a commit that removes the global Observable.

The one downside I discovered to this approach, rather than mandating that your favorite observable library be available as an `Observable` global, is that every observable will need to be wrapped with `Observable.from` or something similar. To see what I mean, try deleting these lines from the test file: https://github.com/acdlite/recompose/blob/fbb62b668ea18d47bb0b6fe8e5503e7b2abbf4d5/src/packages/recompose/__tests__/mapPropsStream-test.js#L20-L29.

I'm thinking maybe on each of the effected modules, we could expose a global module like `configureObservable` that allows you to specify how to convert observables to the desired form:

```js
import configureObservable from 'recompose/configureObservable'
configureObservable(observable => Observable.from(observable)
```

What do you think of that idea?

acdlite:
Pushed an implementation of `configureObservable`. Seems a little gross, but gets the job done.

src/packages/recompose/createEventHandler.js
```diff
@@ -0,0 +1,23 @@
+import $$observable from 'symbol-observable'
+import { createChangeEmitter } from 'change-emitter'
+import { getTransform } from './configureObservable'
+
+const createEventHandler = () => {
+  const emitter = createChangeEmitter()
+  const transform = getTransform()
```
typeetfunc: @acdlite Maybe allow pass to `configureObservable` subject-like anything? And use `change-emitter` if only passed `Observable` dont has method `next`?

chicoxyzzy:
@acdlite I think it's worth it 👍

src/packages/recompose/createEventHandler.js
chicoxyzzy: @typeetfunc `Subject` is Rx-related thing. there are no subjects in most. There is https://github.com/TylorS/most-subject but @TylorS once said in most.js gitter chat that it's more like motorcycle/cycle-related thing.
Discussion about Subjects in most https://github.com/cujojs/most/issues/164

chicoxyzzy:
One thing I can suppose is to add tests for most to be sure in Observable iterop compatibility.

src/packages/recompose/createEventHandler.js
typeetfunc: @chicoxyzzy Thanks for links. Also observer dont has interop point like `symbol-observable` - https://github.com/zenparsing/es-observable/issues/89. Perhaps my proposal was a bit hasty :)

istarkov:
About 2 commits above:
- Import fix is because of `import { Observable } from 'rxjs'` will import all operators.
- Nobind test as a lot of people prefer `rxjs` operators without bind, and test is the best documentation.

chicoxyzzy:
Just trying to implement some most tests (actually an analogue of first one from mapStreamProps) but it seems like `SmartButton` doesn't get mounted by `enzyme` for some reason :confused:

miketembo:
help me understand

acdlite:
@istarkov Please don't add commits to a PR that's not yours without approval. Best way to do that is to submit a PR to this branch :)

What is the value of the test file without bind syntax? I don't see any.

acdlite:
Documentation? I'd rather solve with actual documentation, rather than duplicated tests.

chicoxyzzy:
I spent some time trying to debug my tests until I found this https://github.com/avajs/ava/pull/874
Ughhh... That's why I prefer good old `tape`

miketembo:
please help want to learn recompose eg am trying to show  a spinner below
list how do i do it.

miketembo:
make a rest example with a spinner am looking for an example i can learn
from

chicoxyzzy:
@miketembo this PR is not right place for such questions. Please try [discord](https://discord.gg/0ZcbPKXt5bWAj4rn) or StackOverflow.

typeetfunc:
@acdlite What about `context`? Maybe pass them second arg in componentFromStream? Context doesnt require subscription, becuase its constanst usually. Its maybe useful for pass global streams like as `actions` or `state` streams.

acdlite:
@typeetfunc

```js
const enhance = compose(
  getContext(contextTypes),
  mapPropsStream(props$ => {
    // context is part of props$
  })
)
```

typeetfunc:
@acdlite oh sorry, i missed it. Thanks :)

src/packages/recompose/componentFromStream.js
```diff
@@ -23,17 +32,16 @@ const createComponent = propsToVdom =>
 
     componentWillMount() {
       // Subscribe to child prop changes so we know when to re-render
-      this.subscription = this.vdom$.subscribe(
-        vdom => {
+      this.subscription = this.vdom$.subscribe({
+        next: vdom => {
           this.didReceiveVdom = true
           if (!this.componentHasMounted) {
             this.state = { vdom }
```
istarkov: Why check on hasMounted? (*It was a bug in react, but it solved*)
Also looks like `this.state = {}` could be changed on `this.setState` as this method doe not called from contructor

acdlite: > It was a bug in react, but it solved

Link?

> Also looks like this.state = {} could be changed on this.setState as this method does not called from contructor

You're right. This is left over from when it used to subscribe inside the constructor.

istarkov: > Link ?

One of:
https://github.com/facebook/react/issues/5719
fixed in 15.1

acdlite: The check is to prevent calling `setState` if the component has already unmounted. That issue seems unrelated? Could be wrong.

acdlite:
Re-commenting down here because original comment is on outdated diff:

@istarkov The `componentDidUnmount` check is to prevent calling `setState` if the component has already unmounted. The [issue you linked to](https://github.com/facebook/react/issues/5719) seems unrelated? Could be wrong.

istarkov:
`this.subscription.unsubscribe()` should prevent any calls there, so I deceided that it is because of some bugs. But now I see that in bug case `this.subscription.unsubscribe()` should throw. 

Looks like there is no need to check `this.componentDidUnmount`

acdlite:
Oh duh that's right

chicoxyzzy:
Didn't made any progress with my most.js tests yet and sorry for offtop but I think you may be interested in future of function bind syntax https://github.com/zenparsing/es-function-bind/issues/24#issuecomment-141102543

TL;DR
- it will not be `::`
- probably proposal will be splitted into separate "method extraction" and "functional pipelining" proposals
- it will not be added any soon

src/packages/recompose/componentFromStream.js
```diff
@@ -1,46 +1,43 @@
...
+      [$$observable]() {
+        return this
+      }
+    });
 
     // Stream of vdom
-    vdom$ = propsToVdom(this.props$);
+    vdom$ = toObservable(propsToVdom(this.props$));
 
     didReceiveVdom = false;
```
istarkov: `vdom: null` could me moved into state initializer, so there will be no need in `didReceiveVdom` at all

chicoxyzzy:
Summing up my previous comment I propose to add test which @istarkov provided earlier so no one will suffer refactoring their code after getting inspiration from test files. Also IMO tests should be written in JavaScript, not some dialect no one will understand one time 😃

acdlite:
Yeah I'll remove the `::` operator. I wasn't too concerned about it because it's recommended by the RxJS docs, but I didn't consider that people might read the tests and get confused.

The tests in `mapPropsStream` need to be rewritten anyway, I think. Not too happy with them right now.

acdlite:
@istarkov @chicoxyzzy Just added tests for compatibility with RxJS 4/5, most, xstream, and Bacon. I've also added modules that export the config for each library. Are there any other stream libraries that we should include?

chicoxyzzy:
Wow! Last commit is super cool!

chicoxyzzy:
@acdlite the only library I know besides is Kefir. It's API should be similiar to Bacon

acdlite:
@chicoxyzzy Huzzah, and it already does all the hard work for us! https://rpominov.github.io/kefir/#from-es-observable

acdlite:
I actually Kefir's naming better, too: `fromESObservable` and `toESObservable`, rather than just `fromObservable` and `toObservable`. Much clearer. I'll change it.

PR #207, opened 2016-06-28
Fix mapPropsStream applying transforms twice

acdlite:
Because mapPropsStream composes componentFromStream, the global observable transforms were being applied twice. We'll fix by exporting a factory version of each observable util (including componentFromStream) that allows you opt-out of the global transform.

This was not caught earlier because the tests for mapPropsStream use RxJS 5, which is not affected by double transforms because its observables are ES observables, anyway. To protect against regressions, we'll switch to testing with RxJS4 for now.

Keeping the new factory functions private for now so this can go out as a patch release.


chicoxyzzy:
Do you plan to change API later?

PR #212, opened 2016-06-30
Add babel-register to devDependencies

PeteJodo:
Dumb small pull request but `babel-register` isn't in `package.json` so `npm test` was failing. Added it.


wuct:
Thanks!

wuct:
I also ran into this issue on `node@4`.

PR #221, opened 2016-07-20
Fix typo

maarekj:


istarkov:
Thank you, IMO it's better to change on the name from real definition https://github.com/acdlite/recompose/blob/master/src/packages/recompose/withPropsOnChange.js#L7 `shouldMapOrKeys`

wuct:
`shouldMapOrKeys` is better

maarekj:
It is done.

wuct:
Thanks!
The build has failed because of React unknown-props warning which is not related to this PR.

PR #222, opened 2016-07-21
Update API.md

wallslide:
lodash's omit function takes the object first, and then the array of keys that you want to omit from that object second.


wuct:
The `omit` function mentioned in the doc is the [`lodash/fp`](https://github.com/lodash/lodash/wiki/FP-Guide#lodashfp) version which `keys` is the first argument.

PR #223, opened 2016-07-23
Fix typo

DannyDelott:
`is an stream` -> `is a stream`


wuct:
Thanks!
The build has failed because of React unknown-props warning which is not related to this PR.

PR #225, opened 2016-07-31
Update dependentPropKeys to shouldMapOrKeys in API docs.

andreip:
The signature of `withPropsOnChange` contained the new name `shouldMapOrKeys`, while the explanation referred `dependentPropKeys`.


istarkov:
Thank you!

PR #228, opened 2016-08-11
Add partial and partialProps high order components

Zapix:
Hi all. 

Predefine arguments to passed function to component could be useful in various situations if we need to send token or update special entity. Of course we could use `withHandlers()` for it but to have special high order components looks good for it.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -56,6 +56,7 @@ const PureComponent = pure(BaseComponent)
   + [`lifecycle()`](#lifecycle)
   + [`toClass()`](#toclass)
   + [`partial()`](#partial)
+  + [`partialWithProps()`](#partialWithProps)
 * [Static property helpers](#static-property-helpers)
   + [`setStatic()`](#setstatic)
   + [`setPropTypes()`](#setproptypes)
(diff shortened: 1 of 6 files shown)


istarkov:
Hi, Thank you,
IMO this can be done in a lot of ways in recompose not only with `withHandlers`, 
and as current API allows to do this in a simple way there is no reason to extend-expand API surface. 
It's hard to say NO for really good contribution like this, but the idea of almost all APIs is to keep it as small as possible.

Also I think we should avoid any HOCs with prop names as string parameter like here or in `withState`, 
because these methods could make impossible to use them with typesystems like`flowtype` or `typescript` in the future.

Zapix:
Thank you for your fast response. 

> IMO this can be done in a lot of ways in recompose not only with withHandlers, 
and as current API allows to do this in a simple way there is no reason to extend-expand API surface

When I've looked through api I searched something like [_.partial()](https://lodash.com/docs#partial). It's common behavior and to have this functions on library allow user do not think how to implement them. 

> Also I think we should avoid any HOCs with prop names as string parameter like here or in withState, 
because these methods could make impossible to use them with typesystems like `flowtype` or `typescript` in the future.

I didn't work a lot with typesystems so I don't know is it an issue or not.

istarkov:
IMO all that partial methods was a good idea for old plain javascript, but now for modern js it could be written really short `(...args) => callFn(a, b, ...args)` so no win in readability or anything else.

PR #234, opened 2016-08-20
Update recompose-relay peer dependencies

Secretmapper:
Seems like the peerdeps haven't been updated for a while and the caret flag doesn't work since we're in <1.0.0 (please correct me if I'm wrong)

Thanks!


justinanastos:
This will make my life easier, thanks @Secretmapper  and @acdlite !

istarkov:
Looks like not actual

PR #236, opened 2016-09-05
withVisibilityProp()

r00ger:
The idea behind this is performance improvement in obvious way. If you have components which could be hidden (without unmounting) then you don't want to re-render hidden components.

Even though your component already stateless and pure you still have to implement `shouldComponentUpdate` which isn't convenient (you have to remember about it when you change props of your component).


istarkov:
Thank you!

Isn't this is just the same as next already existing HOC

```javascript
shouldUpdate(
({ visible: prev}, { visible }) => !prev && !visible ,
)
```

Also in your HOC

r00ger:
@istarkov right. Then I think it doesn't make any sense to add such method and make API heavier

istarkov:
Yes, 
I use a lot of internal HOC's like yours in local projects. 
It's easier to think about recompose as about a simple constructor for more heavy HOCs.
Like [here](https://github.com/istarkov/revue/tree/master/src/HOCs), all HOCs use recompose ;-)

r00ger:
I would agree on that. Thanks for fast response

Btw, maybe you can elaborate [Why Higher Order Component is called so?](http://stackoverflow.com/questions/39321027/why-higher-order-component-hoc-is-called-so/39321076)

PR #238, opened 2016-09-08
Update peer dependency for react-relay.

fubhy:
This currently causes shrinkwrap to fail when used with the latest react-relay in the host repository.


fubhy:
Would you be so kind to also consider tagging a new release on npm? :)

fubhy:
Whoops, duplicate of https://github.com/acdlite/recompose/pull/234

PR #248, opened 2016-09-21
Jest snapshot test example (not for merging)

istarkov:
To run

``` bash
npm run jest-test
# OR
npm run jest-test -- --watch
```

I've just played with jest and the first test have found a bug, 
`console.error` in createHelper which was not found with ava ;-)
So +1 to jest.

What are you thinking (everyone who read this) about jest snapshot testing. 

IMO 
`-` tests sources are now less readable - or even unredable. 
`+` BTW test sources with snapshot together is readable for me.
`+` test result output with snapshot diff are more readable for me.

Example:
image

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,7 +3,8 @@
   "parser": "babel-eslint",
   "env": {
     "browser": true,
-    "node": true
+    "node": true,
+    "jest": true,
   },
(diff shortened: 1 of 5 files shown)


wuct:
Nice work!
`+` tests are easier to write

PR #249, opened 2016-09-23
Remove function and arguments length check in createHelper

istarkov:
Failing example
branch caused a console.error at createHelper as functions with default arguments returns incorrect length of arguments.
So [this](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/createHelper.js#L21) was true for branch when all arguments were provided

### DIFF
--- a/src/packages/recompose/__tests__/branch-test.js
+++ b/src/packages/recompose/__tests__/branch-test.js
@@ -1,3 +1,4 @@
+import sinon from 'sinon'
 import test from 'ava'
 import React from 'react'
 import { branch, compose, withState, withProps } from '../'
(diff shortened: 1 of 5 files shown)


istarkov:
Fix

istarkov:
@wuct please review and merge

wuct:
I also ran into this error when I used `createHelper` to build my own HOC before. 

Using `func.length` means we lose the ability to use function default parameters, because `func.length` only includes parameters before the first one with a default value.

IMHO we should just remove the `func.length` checking, since we are not currying by default anymore.

istarkov:
Agree, let's remove that check.

src/packages/recompose/__tests__/branch-test.js
```diff
@@ -49,3 +50,22 @@ test('branch defaults third argument to identity function', t => {
 
   t.is(right, 'Right')
 })
+
+test('branch third argument should not cause console error', t => {
```
istarkov: From one side this test caused this change, from other side this test is not needed now ;-) 
I have no idea what is right solution to remove it now or to allow to stay

istarkov:
Check removed.

src/packages/recompose/__tests__/branch-test.js
wuct: We can just let it stay here :)

wuct:
Nice work!

PR #250, opened 2016-09-27
withReducer: optionally initialize state from default reducer argument

### LINKED ISSUE #247: withReducer should initialize state using default argument
Usually, in Redux reducers can provide initial state from themselves, using default argument value

```js
const counterReducer = (count = 0, action) => {
  switch (action.type) {
    case INCREMENT:
      return count + 1
    default:
      return count
  }
}
```

This reducer works within redux store but doesn't work with recompose as expected.

Is there any chance to fix this? For example, call reducer with arbitrary action to get the initial state only when the last argument (initial state) for `withReducer` is not provided. It will not be breaking for current users, but it will increase reusability of reducers, so you can change them from in-place with recompose to global with redux and the other way round.

just-boris:
fixes #247 

See linked issue for reasons of this change

### DIFF
--- a/src/packages/recompose/withReducer.js
+++ b/src/packages/recompose/withReducer.js
@@ -7,10 +7,17 @@ const withReducer = (stateName, dispatchName, reducer, initialState) =>
     const factory = createEagerFactory(BaseComponent)
     return class extends Component {
       state = {
-        stateValue: typeof initialState === 'function'
-          ? initialState(this.props)
-          : initialState
+        stateValue: this.initalizeStateValue()
       };
+
+      initalizeStateValue() {
+        if (initialState !== undefined) {
+          return typeof initialState === 'function' ?
+            initialState(this.props) :
+            initialState
+        }
+        return reducer(undefined, { type: '@@recompose/INIT' })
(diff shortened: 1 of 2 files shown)


src/packages/recompose/withReducer.js
```diff
@@ -7,11 +7,18 @@ const withReducer = (stateName, dispatchName, reducer, initialState) =>
...
     return class extends Component {
       state = {
-        stateValue: typeof initialState === 'function'
-          ? initialState(this.props)
-          : initialState
+        stateValue: this.initalizeStateValue()
       };
 
+      initalizeStateValue() {
+        if (typeof initialState !== 'undefined') {
```
istarkov: initialState !== undefined is enough, why typeof here?

just-boris: Don't know.
This is just my habit from the time when it was possible to define variable named `undefined`

istarkov:
Thank you!!!

billyjanitsch:
See https://github.com/acdlite/recompose/issues/247#issuecomment-266879979

PR #252, opened 2016-10-03
Add a willMount HOC

abhiaiyer91:
I've been writing small HOCs in my app today that just do something on mount for a dumb component, e.g. analytics calls, SEM tracking, RPC. We're converting our code style to use recompose and was hoping we could contribute this to the project.


istarkov:
Thank you!

1) Recompose already have similar HOC - lifecycle
2) As I know componentWillMount will be deprecated soon https://github.com/facebook/react/issues/7671

abhiaiyer91:
@istarkov Thanks for the reply. We'll just use `lifecycle` then!

Issue #253, opened 2016-10-03
Should `hoistStatics` docs be clarified? 


joekrill:
The API docs for `hoistStatics` simply says:

> Augments a higher-order component so that when used, it copies static properties from the base component to the new component. This is helpful when using Recompose with libraries like Relay.

But under-the-hood it uses `hoist-non-react-statics` -- so any non-react statics _won't_ be hoisted. The docs don't exactly explain (or imply) this. Given that:

1. Should the docs be clarified? I'll create a quick PR for updating the API.md.
2. Is there a legitimate need to hoist statics _including_ non-react statics? For example, being able to access propTypes and defaultProps values?
   * If so, should there be an additional function, or a parameter to `hoistStatics` to ignore specific statics? 
  * If not, should this be renamed to `hoistNonReactStatics` or something more explicit?

istarkov:
1) Merged, let's solve problem one by one
2) I'm sorry I don't know the answer, @wuct @acdlite could you help?

istarkov:
I could just suggest that hoisting React statics will break multiple `defaultProps` and similar enhancers.

wuct:
2) I can't come up with an use case for hoisting non-react statics. In my case, if I want to reuse a `propTypes`, I hoist it manually. IMO this way is easy enough and more clear about what we want to do. For example:

```JavaScript
const itemPropTypes = { ... }

const Item = setPropTypes(itemPropTypes)(...)

const List = setPropTypes({ items: PropTypes.arrayOf(PropTypes.shape(itemProptypes)) })(...)
```

PR #254, opened 2016-10-03
Add details about static types that will not be hoisted with hoistSatics

joekrill:
Clarifies how hoistStatics works, see #253


istarkov:
Super!!! Thank you!

PR #267, opened 2016-10-31
Misc fixes & improvements

matthieuprat:
This is a mixed bag of tiny fixes and improvements that I came up with as I was browsing the docs and the code.


istarkov:
Thank you!!!

PR #272, opened 2016-11-05
add withRefs HOC

belkosyak:
Please consider adding `withRefs` HOC.
I chose to use object-based API and not something like `withRefs('refName', 'refCallbackName')` to support having more than one ref, thats because we need to pass reference to the object with refs as prop to not update base component when prop has been initialised.


istarkov:
Hi, thank you.

Just IMO:

The idea of passing a mutable object through a props looks bad for me. 
`refs` object is a hack, it's mutable, lifecycle of it props depends on component render implementation (what if component will not be rendered or rendered then disabled). 

Also it's not clear in [React documentation](https://facebook.github.io/react/docs/refs-and-the-dom.html) may you or not use refs property with functional components. 

```
You may not use the ref attribute on functional components because they don't have instances. 

You can, however, use the ref attribute inside the render function of a functional component
```

As I know using `ref` in functional component causes it's update every render. And it could be a source of errors, as with using some kind of debounce your refs become outdated. I prefer not to use ref in functional components, and always write a class if I need some kind of `ref` wrapper.

Again IMO refs is a hack and should be used rare and carefully and I don't like the idea to provide enhancers for them.

belkosyak:
Thanks for quick response! Btw I want to take this opportunity and thank you for great set of HOC!
I agree with you, but I think it is matter of compromises you want to take. This HOC worked for me well and I prefer to use functional component over class even in situation when I need refs.
So overall idea was to give users choice which way to pick, as for myself I lacked of this HOC :)
But if you think it can bring more bad than good, I also can understand this :)
Have a great day!

belkosyak:
And one more thing:
>As I know using ref in functional component causes it's update every render. And it could be a source of errors, as with using some kind of debounce your refs become outdated.

I checked up with React 15.1.0 and it looks like even if you use class `ref` callback also fires every render. However docs say:
>React will call the ref callback with the DOM element when the component mounts, and call it with null when it unmounts.
>
>Using the ref callback just to set a property on the class is a common pattern for accessing DOM elements. If you are currently using this.refs.myRefName to access refs, we recommend using this pattern instead.

istarkov:
I understand all that cases above, I wrote that in a case with using `ref`s without mutable object.
There will be not such problem with your code. 
BTW having my theory that 
 - passing mutable objects is a hack, 
 - refs is hack too, 
 - refs with functional components is double hack 

so there are more hacks than my mind could accept ;-) 

For me and my code (not sure that it's for all) `refs` is very rare used feature,
and I use refs mostly in low level controls 
and just found one case where I use refs with `react-virtualized`

BTW it was done with mapPropsStream

``` javascript 
mapPropsStream((props$) => {
  let ref_;
  const registerChild = (ref) => ref_=ref;
  // use ref_ only here no refs pass to children
  .....
  return blabla$.combineBlbalala(..., {
    props,
    registerChild,
  });
})
```

belkosyak:
Thanks for the detailed response! I agree with you on every point.
In my code I use refs pretty frequently, but it is because we implement tabs functionality inside app and because of it we need to restore scroll when user switches between tabs, but you right it is rare case in general.
But looks like I need to rethink my code to avoid problems in future, since hacks always has some pitfalls :)))
Thanks for time you spent! I think there is no reason to keep PR open anymore! I hope this conversation will helpful for someone at least :)
Have a nice day!

SeeThruHead:
@belkosyak if you make this a standalone addition to recompose it would get plenty of use.

belkosyak:
@SeeThruHead yeah maybe you right, I will give it a shot :)

istarkov:
#295 somehow can be helpful, I'll close this for now

PR #275, opened 2016-11-11
Update isReferentiallyTransparentFunctionComponent.js

oliviertassinari:
I'm wondering about the purpose of `!Component.propTypes`.
Using `propTypes` is quite a popular pattern.
If it's only here in order to get proper warning in dev mode, that shouldn't prevent us from
instantiation less react components in production.

Hopefully, we can use [babel-plugin-transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) to increase the number of referentially transparent components

### DIFF
--- a/src/packages/recompose/isReferentiallyTransparentFunctionComponent.js
+++ b/src/packages/recompose/isReferentiallyTransparentFunctionComponent.js
@@ -5,7 +5,7 @@ const isReferentiallyTransparentFunctionComponent = Component => Boolean(
   !isClassComponent(Component) &&
   !Component.defaultProps &&
   !Component.contextTypes &&
-  !Component.propTypes
+  (process.env.NODE_ENV === 'production' || !Component.propTypes)
 )
 
 export default isReferentiallyTransparentFunctionComponent


wuct:
If we use the babel plugin to remove `propTypes` in production, there is no need to check the environment here again, right?

oliviertassinari:
> If we use the babel plugin to remove propTypes in production

- Removing 100% of the `propTypes` in production is quite challenging. You need either lib author do wrap there `propTypes` with a `process.env.NODE_ENV === 'production'` or make babel parse the all `node_modules` folder. Not that easy.
- It's a micro optimization, but even if you remove 100% of the `propTypes`, I think that checking `!Component.propTypes` is a lost CPU cycle in production.

wuct:
I get your point now.  IMO `!Component.propTypes` should be removed in production, because React doesn't check `propTypes` in production.

src/packages/recompose/isReferentiallyTransparentFunctionComponent.js
```diff
@@ -5,7 +5,7 @@ const isReferentiallyTransparentFunctionComponent = Component => Boolean(
   !isClassComponent(Component) &&
   !Component.defaultProps &&
   !Component.contextTypes &&
-  !Component.propTypes
+  (!Component.propTypes || process.env.NODE_ENV === 'production')
```
wuct: Will this line be considered dead code and be eliminated in production?

istarkov: swapping on `process.env.NODE_ENV === 'production' || !Component.propTypes` will avoid getting the `propTypes`

oliviertassinari: @istarkov I agree that it would be better. I believe that uglify-js is smart enough to remove those `if (!foo.bar || true)` expression but that may not be true with other minifier, they may keep it.

wuct: Yes, I've just tried [UglifyJS](https://skalman.github.io/UglifyJS-online/), both `true || foo` and `foo || true` work.

wuct:
LGTM. I'd like to know how @acdlite and @istarkov think.

istarkov:
I like this PR 👍 to merge

wuct:
@oliviertassinari thanks!

oliviertassinari:
Awesome, thanks 🎉 !

Any plan to do a release? The last one was *144 days* ago.
We use the decoration pattern quite extensively at @doctolib.
Performance is important for us. @neoziro is even investigating toward combining all those HoCs into a single one with [recompact](https://github.com/neoziro/recompact) while keeping the same API.

neoziro:
I agree with @oliviertassinari, the project is moving slowly, that's why we are building "recompact" apart recompose. We will switch to it and use it in production. If the experimentation is a success, we could consider merging the two projects in the future.

PR #276, opened 2016-11-13
Export the root object

wuct:
Instead of importing functions individually

```JS
import { compose, mapProps, pure, withHandlers ... } from 'recompose'
```

we can just import the root object

```JS
import Rc from 'recompose'
```

One of the benefit of importing the root object is preserving the namespace, which is useful when we use `recompose` with other functional libraries.

Considering optimizing the bundle size, we can still use `babel-lodash-plugin`.


nfcampos:
curious, what's the advantage for users to do 
```js
import Rc from 'recompose'
```
instead of
```js
import * as Rc from 'recompose'
```
?

wuct:
@nfcampos both works similarly. As you can see, I also use `import * as ...` [here](https://github.com/acdlite/recompose/blob/dacf8d7068eeeb907ce28ead477bdb1f592ca91d/src/packages/recompose/index.js#L1).

IMHO `import Rc from 'recompose` is more concise and kind of a convention now.

wuct:
I am getting used to `import * as ...` now, so I am going to close this issue.

Issue #259, opened 2016-10-19
withState update initial properties?


macrozone:
`withState` accepts a function to return default state:

```
withState('show', 'setShow', ({ showInitally }) => showInitally);
```

so if the resulting component is called with `<MyComponent showInitally />`
the `show` state will be true.

However, if props get updated (showInitally changes to false), the state will not update.

How can you achieve that state changes when props change?

istarkov:
For logic where state could have a multiple sources of true I highly recommend you to use 
`mapPropsStream` enhancer, with rxjs or any other library. 
Or to write your own enhancer.

In most cases multiple sources of one state is a bad idea, but yes there are few situations we could need it. 
I used such logic a lot in the past, but now every time I see the need of updating internal state based on external properties, I try to find a solution how to rewrite code and get just one source of true.

istarkov:
I'll try explain shortly the problems I got trying to write universal solution of issue above.

What if internal state was changed inside component, 
then component props changed, do I need to reset state or not. 
Do I need to reset state if some props changed, or on every render.
How I could reset state to initial value if props it depends were not changed.

I tried a lot of solutions, and all of them had issues, I even wrote issues about https://github.com/acdlite/recompose/issues/199

And be sure example in #199 has bugs.

wuct:
I am agree with @istarkov's suggestion. However, if what you need is a quick hack, you can change `key` to force React to remount the component and re-evaluate the initial state. Ex:

```JavaScript
<MyComponent showInitally key={showInitally} />
```

Please know that this approach is a hack and should be used with caution.

istarkov:
@wuct may be not a hack as sometimes it's needed to force creating a new DOM element.

For example I use it to force creating new img element on src chage `<img src={src} key={src} />`  Because otherwise until new image loaded you will see previous image, and in a lot of cases it's not expected behaviour.

wuct:
`<img src={src} key={src} />` is a great use case!

macrozone:
Hi Guys,

thanks for your answers!

Maybe some more background infos: i use withState together with mantra specification: https://github.com/kadirahq/mantra in a meteor-app.

mantra also uses react-komposer which also uses HOCs to compose containers. See a sample for meteor here: https://github.com/kadirahq/react-komposer#using-with-meteor

In the sample above onData is called with new "posts" whenever posts change. posts is a property that gets passed to the next component. So you can do something like this (slightly altered sample):

```

function composer({postId}, onData) {
[...]

Issue #199, opened 2016-06-13
withStateSelector proposal


istarkov:
Idea is to provide withState with *resetable* state. 
Sometimes when I work with `react-router` (not only) I don't want to recreate component to reset state after route has changed. 
The idea is to provide `withStateSelector(stateName, stateUpdaterName, selectorFactory)` HOC which will reset - recalculate `stateValue` every time selector return a new result.

This will be easy to use with [reselect](https://github.com/reactjs/reselect#createselectorinputselectors--inputselectors-resultfunc) library. Like
 ```javascript
  withStateSelector(
    'markers', 'setMarkers',
    () => createSelector(
      ({ route: { markersCount = 20 } }) => markersCount,
      (markersCount) => generateMarkers(markersCount)
    )
  ), 
```

The real code with example:

[Usage](https://github.com/istarkov/google-map-react/blob/master/develop/GMapOptim.js#L62-L69)
[Routes with params](https://github.com/istarkov/google-map-react/blob/master/develop/Main.js#L17-L19)
[withStateSelector source](https://github.com/istarkov/google-map-react/blob/master/develop/utils/withStateSelector.js)

Here every time route changed and got new `markersCount` param `createSelector` will return a new result, and state will be resetted.

PS: It's useful with controls like forms, which should use internal state until save 'button' / 'blur event' is pressed / occured.

istarkov:
Also we could just extend withState - and to detect `selectors` if result of initial state call is function.

istarkov:
Also to make it more same with `reselect` it's possible to pass to selector also the current `state`,
`this.selector(this.state.stateValue, nextProps)`

jurgob:
+1 I need it too

PR #278, opened 2016-11-22
[POC] bind state to props

vaukalak:
Hello, this is just a draft proposal of how we could fix https://github.com/acdlite/recompose/issues/259.

### DIFF
--- a/src/packages/recompose/withState.js
+++ b/src/packages/recompose/withState.js
@@ -2,7 +2,7 @@ import { Component } from 'react'
 import createHelper from './createHelper'
 import createEagerFactory from './createEagerFactory'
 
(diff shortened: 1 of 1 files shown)


istarkov:
Hi, 
the main problem of this HOCs that `state` will be dependent not on data but on `render` calls.
It's somehow similar to #199 and even with selector it has the same problem.

The other way is to use `initialState` function with signature like `(nextProps, props, state)` similar to all that we have in `componentWillReceiveProps`.

I wrote here https://github.com/acdlite/recompose/issues/259#issuecomment-254812353
that having multiple source of one state is not a good idea, and it's better to avoid such situations, or solve with tools like `rx`.

PR #280, opened 2016-11-28
Add link to the createClass spec in the lifecycle docs

chrismendis:
It's surprisingly difficult to glean the exact specification of createClass from the React docs. This presents some difficulties with figuring out what exactly one can do with the `lifecycle` function.

Perhaps we can at least add a link to the actual spec in the React code?  This is admittedly a bit fragile, but I can't think of a better solution off the top of my head.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -451,7 +451,7 @@ lifecycle(
 ): HigherOrderComponent
 ```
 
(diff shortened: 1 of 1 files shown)


istarkov:
Hi thank you, 

The link you've added is not stable, any change to `ReactClass.js` could make this link broken.

I agree that current React documentation make it difficult to understand what is `specification` object without links to https://facebook.github.io/react/docs/react-without-es6.html etc.

BTW I think in the future `createClass` will be deprecated, so giving a link to `specification` object with all that `mixins` property can be bad idea ;-)

For now I think that `lifecycle` enhancer is also bad idea ;-), I almost removed it from my code. As in all places I used it, it was better to use usual `es6` class for component itself.

chrismendis:
@istarkov sounds good! I also had to abandon using `lifecycle` because I needed to use refs.

PR #290, opened 2016-12-13
Update symbol-observable to 1.0

### LINKED ISSUE #246: Make symbol-observable a peer dependency or update it to @latest
Right now recompose have a dependency on "symbol-observable": "^0.2.4" (not peer but direct) so I end up with this dependencies. 

recompose@0.20.2
--------symbol-observable@0.2.4
rxjs@5.0.0-beta.11
symbol-observable@1.0.2

I don't know much on how rxjs depends on symbol-observable, but certainly having two symbol-observable libraries is not a good thing.

danez:
Version 1.0 of symbol-observable is only breaking if it would be used with `require('symbol-observable')`, as recompose already uses es6 it can be upgraded without running into this breaking change as babel is already taking care about the default export.

All other changes should not affect recompose.

Here is the changelog:
 https://github.com/blesh/symbol-observable/blob/master/CHANGELOG.md

IT would be nice to do this, as currently having redux and recompose in a project, results in two versions of `symbol-observable` being bundled.

Fixes #246

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "gzip-size-cli": "^1.0.0",
     "jsdom": "^8.4.0",
     "kefir": "^3.2.3",
-    "most": "^0.19.7",
+    "most": "^1.0.2",
     "nyc": "^6.4.0",
     "react": "^15.0.0",
     "react-addons-test-utils": "^15.0.0",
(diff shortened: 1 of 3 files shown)


istarkov:
Looks like `most` version also need to be updated?

danez:
Yes noticed that. I made recompose compatible with 0.19 and 1.0 of most (theoretically) as the exports changed.

Only problem still is, that users of most 0.19 although the correct export is used can't use it now, as symbol-observable 1.0 is not compatible with most < 1.0.2

istarkov:
Thank you!!!

PR #293, opened 2016-12-13
Simplified the composer to use higher order components

JoeCortopassi:
*higher order functions


istarkov:
I swear I saw in the past rewriting compose function from `reduce` version into "plain `for`" cycle in order of some performance benefits ;-)

I like this, looks more readable.

istarkov:
Thank you!!!

istarkov:
Comment about performance of this https://github.com/reactjs/redux/pull/2145#issuecomment-266264790

PR #295, opened 2016-12-23
Allow handlers property of withHandlers to be a factory functions

istarkov:
Partially the same reason as `mapStateToProps` in react-redux  allows return a factory. 

To have ability to use local selectors see [reselect](https://github.com/reactjs/reselect) (memoization, cache) 

Like 

```javascript
withHandlers((/* initialProps */) => {
  const selector_ = createSelector(({ x }) => x, x => someCalculus(x));
  return {
    getMemoizedResult: () => selector_,
  };
})
```

This is somehow similar to `withPropsOnChange` but here you can use memoized data on demand.

The other case is to hold internal state which must not cause `setState` calls, like references, for which currently people try to use different hacks with recompose. (In my case `scrollTop` from scroll events) 

```javascript
withHandlers((/* initialProps */) => {
  let ref_;
  // let handle_;
  return {
    registerChild: () => (ref) => {
       ref_ = ref;
       // if(ref_ === null) clearTimeout(handle_);
    },
    onResize: () => () => {
      ref_.updateLayout();
      // handle_ = setTimeout(ref_.updateLayout, 0); // bad hack to not use didUpdate
    }
  };
})
```

@wuct, others your thoughts?

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -130,11 +130,14 @@ Instead of an array of prop keys, the first parameter can also be a function tha
 withHandlers(
   handlerCreators: {
     [handlerName: string]: (props: Object) => Function
+  } |
+  handlerCreatorsFactory: (initialProps) => {
+    [handlerName: string]: (props: Object) => Function
   }
 ): HigherOrderComponent
 ```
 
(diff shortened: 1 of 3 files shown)


wuct:
Great idea! I am all for it.

PR #298, opened 2016-12-24
Remove all Unknown prop warning from tests

istarkov:
Few screens of tests warning like

```
Warning: Unknown prop `blabla` on <blublu> tag. 
Remove this prop from the element. 
For details, see https://fb.me/react-unknown-prop
```

has fixed (removed).

Ava updated.


wuct:
Nice work!

Issue #124, opened 2016-04-06
Prune list of helpers


acdlite:
As we start thinking about a 1.0 release, I want to drop any helpers that haven't worked out or are out of scope. Let's use this issue to keep track of which ones are on the chopping block:

- [x] `withAttachedProps` — will be replaced by `withHandlers` (~~I'll have PR ready for this soon~~ added in 0.17)
- [x] ~~`mapPropsOnChange` — not convinced this is actually useful. Can be replaced by normal `mapProps` + function memoization, via lodash or otherwise.~~ Keeping this and renaming to `withPropsOnChange()` https://github.com/acdlite/recompose/pull/134
- [ ] `doOnReceiveProps`
- [ ] `lifecycle`

Update: there's some disagreement over whether the lifecycle-based helpers belong in the library or not. My attitude is that they're currently half-baked, so let's remove them for now and we can add them in a minor release once we've figured out a better story for them — or not, if we decide they're outside the scope of the library.

istarkov:
Without `lifecycle` it will be impossible to cancel async tasks.

istarkov:
BTF I'm not using `lifecycle` directly, and use my cancelAsync HOC based on lifecycle, which can be easily be rewritten as simple component.

nfcampos:
`doOnReceiveProps` is pretty useful, especially as a building block for other things, it's basically createSink but you can also render something

acdlite:
My thinking is that both `lifecycle` and `doOnReceiveProps` are too leaky. @nfcampos Can you give me an example of how you've used `doOnReceiveProps`?

RafalFilipek:
+1 for `withAttachedProps` & `mapPropsOnChange`
no opinion - `doOnReceiveProps`
`lifecycle` - should we use classes for that?

acdlite:
@RafalFilipek Yes, or something like rx-recompose's `observeProps`

nfcampos:
@acdlite actually you're right, it's not that useful, at least for me, I thought I remembered two times I had used it but going back to find them I now realise that
- in one it was just a pretty stupid use so nevermind ;)
- in the other I thought about using it but ended up not using it because I also needed access to the old props inside the callback (to compare) (this 2nd use is this https://gist.github.com/nfcampos/512076429479c7c9b7243be0b33ec924 a HOC like `withProps` but for props that are computed asynchronously, if you think something like that belongs in `recompose` I'd be more than happy to contribute)

istarkov:
[...]

PR #300, opened 2017-01-04
Added recursive pick for withPropsOnChange feature

EloB:
Hi

First of all I would like to thank you for this awesome library!

I often use the `withPropsOnChange` but I miss the functionality to listen for nested properties and writing functions looks totally ugly. So I came up with this.

```js
withPropsOnChange(
  (props, nextProps) =>
    props.router.routeParams.something !== nextProps.router.routeParams.something ||
    props.somethingElse !== nextProps.somethingElse ||
    props.thisStartToLookUgly !== nextProps.thisStartToLookUgly,
  ({
    fetchSomething,
    router: { routeParams: { something } },
    somethingElse,
    thisStartToLookUgly
  }) => fetchSomething(something, somethingElse, thisStartToLookUgly)
)
```

The code looks even more ugly if the `routeParams` isn't defined with current implementation. You then need to have checks if it exists that bloats the code.

With my feature request you can replace this with:

```js
withPropsOnChange(
  [
    ['router', 'routeParams', 'something'],
    'somethingElse',
    'thisStartToLookGood'
  ],
  ({
    fetchSomething,
    router: { routeParams: { something } },
    somethingElse,
    thisStartToLookGood
  }) => fetchSomething(something, somethingElse, thisStartToLookGood)
)
```

### DIFF
--- a/src/packages/recompose/__tests__/pick-test.js
+++ b/src/packages/recompose/__tests__/pick-test.js
@@ -1,0 +1,29 @@
+import test from 'ava'
+import pick from '../utils/pick'
+
+test('pick', t => {
+  t.deepEqual(
+    pick({
+      one: '1',
+      dont: 'include',
+      two: {
+        deep: {
+          value: 'true'
+        }
+      },
+      root: {},
+      three: '3'
+    }, [
+      'one',
+      ['two', 'deep', 'value'],
+      ['root', 'missed'],
+      'three'
+    ]),
+    {
+      one: '1',
+      'two\ndeep\nvalue': 'true',
+      'root\nmissed': undefined,
+      three: '3'
+    }
+  )
+})
--- a/src/packages/recompose/utils/pick.js
+++ b/src/packages/recompose/utils/pick.js
@@ -2,7 +2,18 @@ const pick = (obj, keys) => {
   const result = {}
   for (let i = 0; i < keys.length; i++) {
     const key = keys[i]
-    if (obj.hasOwnProperty(key)) {
+    if (Array.isArray(key)) {
+      let value = obj
+      for (let x = 0; x < key.length; x++) {
+        const deepKey = key[x]
+        if (!value.hasOwnProperty(deepKey)) {
+          value = undefined
+          break
+        }
+        value = value[deepKey]
+      }
+      result[key.join('\n')] = value
+    } else if (obj.hasOwnProperty(key)) {
       result[key] = obj[key]
     }
   }


src/packages/recompose/__tests__/pick-test.js
```diff
@@ -0,0 +1,29 @@
...
+      three: '3'
+    }, [
+      'one',
+      ['two', 'deep', 'value'],
+      ['root', 'missed'],
+      'three'
+    ]),
+    {
+      one: '1',
+      'two\ndeep\nvalue': 'true',
```
EloB: Seperator could be changed if we don't like the new line but property names probably won't include new lines (that is why I choose that character).

src/packages/recompose/__tests__/pick-test.js
```diff
@@ -0,0 +1,29 @@
...
+        deep: {
+          value: 'true'
+        }
+      },
+      root: {},
+      three: '3'
+    }, [
+      'one',
+      ['two', 'deep', 'value'],
+      ['root', 'missed'],
```
EloB: If we try to go deeper `['root', 'missed', 'something']` would still return `undefined`.

src/packages/recompose/__tests__/pick-test.js
EloB: Also if we don't like this solution we could do this with a `Map` but then we have to add a new `shallowEqual` that supports `Map`.

istarkov:
Hi, thank you!
Can't say strongly no, but

Why not just use something like 
`withProps(({ router: { routeParams: { something } } }) => ({ something })),`
before `withPropsOnChange(` it does not give a lot of overhead in number of code lines, or readability.

Also,  I think that even I was one of the persons who saved this enhancer [comment](https://github.com/acdlite/recompose/issues/124#issuecomment-207749189) it's mostly overused. 

I used it in every component I did, but now it's very rare enhancer in my code, javascript and React are fast enough to use it only in real "computational intensive" tasks. (Can say that such tasks are very rare in our codebase)

Also I dislike all that string getters in the code like `get('x.y.z', a)` or like in current recompose code, yes it makes code shorter, but in the near nice future (IMO ;-)) it will make impossible to use all that great features we see in editors of languages like `C#`, `Java` etc - I mean code refactoring (rename property), find all property usage, etc...

So from my POV we should not create or extend features non possible in strongly typed languages.

PS:
Also I wanna say if you use it for calling some asynchronous actions inside - not sure what means  `fetchSomething` it's not the right place to do it, as there are few problems here - no cancellation provided, double renders, non sync state etc. 

In my code I very rare use version of `get` similar to this [gist](https://gist.github.com/istarkov/0ba94abff98db07c19efdc88ebd350a2)

And write `pick like` constructions using modern `js` destructuring.

EloB:
```js
withProps(({ router: { routeParams: { something } } }) => ({ something }))
```
Will fail if either `router` or `routeParams`. Sorry for short comment written on mobile 😁

istarkov:
Will not fail as you can use defaults for args

EloB:
So I need 3 types of hocs then? That I will also call bloating the code 😂

wuct:
The current signature of `pick` is same as  [`lodash.pick`](https://lodash.com/docs/4.17.0#pick) and [`ramda.pick`](http://ramdajs.com/0.21.0/docs/#pick), so I think changing its signature here is no go.

> So I need 3 types of hocs then? That I will also call bloating the code 😂

I believe you meant using `defaultProps`. I seldom use it now. Instead, I use [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) or [ramda.path](http://ramdajs.com/0.21.0/docs/#pathOr) a lot.  I find this approach makes my HOCs more readable. So, in your case, 3 HOCs can be reduce to 2 HOCs.

If it's still not good enough to you, one of the beauty of FP is that you can always compose a new function easily. Without changing the behavior of `withPropsOnChange`, you can create a function with a signature `(keys: string[][]) -> (props: Object, nextProps: Object) -> bool` and use it with `withPropsOnChange`.

EloB:
@wuct So you also want another dependency? I don't like to have 1 000 000 dependencies in my `package.json` file. Feels like `left-pad`.

Default parameters in this case that is very common looks totally ugly.
```js
withPropsOnChange(
  ({ router: { params: { something } = {} } = {} }) => {

  }
)
```

EloB:
@istarkov About that fetch claim. Their are other ways of solving that.
For example https://github.com/plougsgaard/react-timeout

bertho-zero:
Why pick does not support dot notation like lodash? it would be useful for onlyUpdateForKeys for example.

PR #302, opened 2017-01-09
Add toESObservable identity fn to rxjsObservableConfig

trxcllnt:
This PR adds a `toESObservable` function to the `rxjsObservableConfig`, so it can be used with `componentFromStreamWithConfig`. I ran into this issue writing a library with recompose and RxJS. I don't want the library overwrite if someone has set the global config, but using `rxjsObservableConfig` with `componentFromStreamWithConfig` will throw an error, because `toESObservable` isn't defined.

It seems the tests for `componentFromStream` are skipped, so I didn't add a test for this case. Let me know if there's anything else I should add to this PR. Thanks.

### DIFF
--- a/src/packages/recompose/rxjsObservableConfig.js
+++ b/src/packages/recompose/rxjsObservableConfig.js
@@ -1,7 +1,8 @@
 import Rx from 'rxjs'
 
 const config = {
-  fromESObservable: Rx.Observable.from
+  fromESObservable: Rx.Observable.from,
+  toESObservable: stream => stream
 }
 
 export default config


wuct:
It makes sense to me, but I want to clarify your requirement. 

When it comes to global config, do you mean the config in [`setObservableConfig`](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/setObservableConfig.js)? If the answer is yes, why don't you import the config from it? If no, what does the global config mean?

trxcllnt:
@wuct Yep, exactly. As I'm implementing a library that uses both recompose and Rx, I figured it's probably best not to call `setObservableConfig`, as someone who imports my library fn may have already set the global config to something they prefer. Here's an example of the code I figured I'd write:

```es6
// my-lib/hoc.js
import { Observable } from 'rxjs/Observable';
import rxjsObservableConfig from 'recompose/rxjsObservableConfig';
import { mapPropsStreamWithConfig } from 'recompose/mapPropsStream';

const mapPropsStream = mapPropsStreamWithConfig(rxjsObservableConfig);
const myHOC = mapPropsStream((props) => /* etc. */);

export { myHOC };
export default myHOC;
```

wuct:
I see. Thanks for providing detail.

PR #303, opened 2017-01-09
Сomplete props stream before unmounting.

### LINKED ISSUE #220: [rx] props stream never ends
When `ComponentFromStream` is unmounted, it's obvious that `props$` stream will no more receive new values – i.e., it's ended.
So, maybe, it makes sense to call `observer.complete()` in `componentWillUnmount`?
It can simplify performing cleanup on unmount, without need to use `lifecycle` helper and implicitly convert our functional components into classes.

Before:

``` js
compose(
   mapPropsStream(props$ => {
       ...setup some handlers and cleanup function...
       return props$.combineLatest(...);
   }),
   lifecycle({
      componentWillUnmount() {
          this.props.doSomeCleanup();
      }
   }),
   MyFunctionalComponent
)
```

After:

``` js
componentFromStream(props$ => {
    ...setup some handlers and cleanup function...
    props$.doOnCompleted(doSomeCleanup);
    return props$.combineLatest(..., MyFunctionalComponent);
})
```

umidbekkarimov:
Closes #220

### DIFF
--- a/src/packages/recompose/__tests__/componentFromStream-test.skip.js
+++ b/src/packages/recompose/__tests__/componentFromStream-test.skip.js
@@ -62,3 +62,35 @@ test('handler multiple observers of props stream', t => {
   wrapper.setProps({ value: 2 })
   t.is(div.prop('value'), 2)
 })
+
+test('complete props stream before unmounting', t => {
+  let counter = 0
+
+  const Div = componentFromStream(props$ => {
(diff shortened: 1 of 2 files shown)


istarkov:
Plz see comment in #220 about second argument

umidbekkarimov:
@istarkov done

istarkov:
Thank you!

PR #306, opened 2017-01-27
Change string component creation using React.createElement in toClass function

mattiaocchiuto:
Using the `toClass` function to create a component using a string it create a component with the `baseComponent` variable not resolved, so if I try to create a component calling: `toClass('textarea')` I will receive as output a `<baseComponent>` element and not a `<textarea>` element.


istarkov:
Thank you, good catch, please fix linting errors

mattiaocchiuto:
Done

PR #309, opened 2017-01-29
componentFromStream can potentially throw EmptyError: no elements in sequence

umidbekkarimov:
As from 0.22.0 `componentFromStream` will complete the stream during unmount and usage of [first](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-first), [last](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-last) or [single](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-single) methods can potentially throw [EmptyError](http://reactivex.io/rxjs/class/es6/util/EmptyError.js~EmptyError.html).

See #220 and #303

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -702,6 +702,8 @@ const Counter = componentFromStream(props$ => {
 })
 ```
 
+**Note: As from 0.22.0 componentFromStream will complete the stream and usage of [first](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-first), [last](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-last) or [single](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-single) methods can potentially throw [EmptyError](http://reactivex.io/rxjs/class/es6/util/EmptyError.js~EmptyError.html).**
+
 ### `mapPropsStream()`
 
 ```js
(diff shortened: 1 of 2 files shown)


istarkov:
are u sure https://github.com/acdlite/recompose/blob/master/src/packages/recompose/componentFromStream.js#L41
any example with real life error throwing?

umidbekkarimov:
Had this problem in my project after update. It was cause with `propsStream.filter(props => props.id > 0).first()` like streams. I'll add test cases bit later.

istarkov:
Ill add a ref to this thread here https://github.com/acdlite/recompose/releases/tag/v0.22.0 
dont think that documentation is the place for possible issues

istarkov:
Done, thank you.

umidbekkarimov:
> any example with real life error throwing?

Done

> dont think that documentation is the place for possible issues

I thought it would be good for new users to know about this gotcha. And yes, for old users release description is better place.

istarkov:
May we move changes from documentation into first comment here and Ill merge the test only?

istarkov:
I think for new users it will be well described understandable rx error, isnt it?

istarkov:
Or may be change documentation as now its not obvious why such exception should occur

umidbekkarimov:
> May we move changes from documentation into first comment here and Ill merge the test only?

Done

> Or may be change documentation as now its not obvious why such exception should occur

I'm not good at documenting things :).

istarkov:
A rare persons are good in documentation. My own projects some kind of documentation hell ;-)

istarkov:
Thank you!!!

Issue #54, opened 2015-11-12
request: Add HOC to support "get async data on mount" pattern.


jimbolla:
I have this pattern all over my codebase:

    const ComponentA = React.createClass({

        propTypes: {
            id: PropTypes.string,
        },

        contextTypes,

        getInitialState() {
            return {};
        },

        componentDidMount() {
            http.get(this.context.rootUrl + "SomeAPI", {id: this.props.id})
                .then(config => {
                    if (this.isMounted())
                        this.setState({data});
                });
        },

        render() {
            return this.state.data
                ? (<ComponentB {...this.props} {...this.state.data}/>)
                : (<Spinner/>);
        }
    });

A component, such as ComponentA, has to make an ajax call on mount, and then pass that data to another component on completion. I'm imagining a HOC signature that looks something like this:

    onMount(
        getData: (props: Object, context: Object) => Promise,
        BaseComponent: ReactElementType
    ): ReactElementType

acdlite:
Thanks for the suggestion! I'm not sure to what extent Recompose should be providing lifecycle hooks. I'd like to think about this for a while before making any decisions.

jimbolla:
Hmmm. Well, one option would be to have a separate package `recompose-lifecycle` for stuff like that, if there's enough code to justify one. Anyways, here's my naive implementation of onMount:

    const onMount = (getData, BaseComponent) => React.createClass({

        contextTypes: BaseComponent.contextTypes,

        getInitialState() {
            return {};
        },

        componentDidMount() {
            getData(this.props, this.context)
                .then(promise => {
                    if (this.isMounted)
                        this.setState({promise});
                });
        },

        render() {
            return (
                <BaseComponent {...this.props} {...this.state.promise} />
            );
        },
    });

RafalFilipek:
Lifecycle events are cool but I think there is no reason for packages like `recompose-lifecycle` in that case. Quick example:

````
import React from 'react'
import { render } from 'react-dom'
import { compose, lifecycle, branch, renderNothing, withState, renderComponent } from 'recompose'

// returns HOC
const componentWithApi = (BaseComponent, requests) => (
  compose(
    // are we done?
    withState('loaded', 'setLoaded', false),
    lifecycle(
      (c) => (
        // Lets make all requests and then update state
        Promise.all(requests.map(r => r()))
        .then(() => c.props.setLoaded(true))
      ),
      () => {}
    ),
    // You can show some loader while data are loading or
    // render nothig.
    branch(
[...]

PR #310, opened 2017-01-30
add note for the use of setState in lifecycle

spoike:
Adding a note to clear things up with using state in `lifecycle`.

I was trying to implement an async data loading component and the doc doesn't say anything about how `this.state` is handled which is confusing. So I had to read the source code to determine if state changes were propagated. As suspected it does and that fact should be included in the docs.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -456,6 +456,8 @@ lifecycle(
 
 A higher-order component version of [`React.createClass()`](https://facebook.github.io/react/docs/react-api.html#createclass). It supports the entire `createClass()` API, except the `render()` method, which is implemented by default (and overridden if specified; an error will be logged to the console). You should use this helper as an escape hatch, in case you need to access component lifecycle methods.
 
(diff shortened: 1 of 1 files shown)


istarkov:
Thank you, but isnt it obvious?

spoike:
@istarkov Being new to `recompose` it wasn't obvious for me as there are no concrete examples and no mention of the usage of state anywhere.

What led me to create this PR was https://github.com/acdlite/recompose/issues/54#issuecomment-213566088 that doesn't use `this.state`; and also neither does the README for [https://www.npmjs.com/package/react-lifecycle-hoc](react-lifecycle-hoc) mention this aspect of it either. I believe this change will make it less confusing for newcomers.

istarkov:
I see, 
PS:
want to say that in nearest React version, createClass will be deprecated and we need to decide what to do with lifecycle, IMO it should be deprecated in recompose also but it's just IMO.

Issue #287, opened 2016-12-12
Docs broken about branch


nemocurcic:
The docs contains this:
```js
// `hasLoaded()` is a function that returns whether or not the component
// has all the props it needs
const spinnerWhileLoading = hasLoaded =>
  branch(
    props => !hasLoaded(props),
    renderComponent(Spinner) // `Spinner` is a React component
  )

// Now use the `spinnerWhileLoading()` helper to add a loading spinner to any
// base component
const enhance = spinnerWhileLoading(
  props => props.title && props.author && props.content
)
const Post = enhance(({ title, author, content }) =>
  <article>
    <h1>{title}</h1>
    <h2>By {author.name}</h2>
    <div>{content}</div>
  </article>
)
```

But `branch` doesn't support only 2 arguments. I saw that there were a PR about this but it's not merged. https://github.com/acdlite/recompose/pull/227

Please publish the new functionality or change the docs!

istarkov:
Merged, 
![image](https://cloud.githubusercontent.com/assets/5077042/21113298/09df0a9a-c0ba-11e6-9e48-5f07c98c2856.png)
not published yet.

Documentation updated according to PRs, so the problem in releases,
having that only @acdlite can publish this repo, I can't say much when, why, who.

istarkov:
PS: Until release just use latest tag for the documentation
https://github.com/acdlite/recompose/tree/v0.20.2

istarkov:
Oops, 
now I can publish on npm, 
so new version available on npm 0.21.0, please check.

;-)

PR #311, opened 2017-01-30
update doc on branch on optional parameter

spoike:
The docs for `branch` does not mention the fact the the third argument is optional or that the function is kinda curried. This is not inferred until you read the example in `renderComponent` which is contextually the not really the place you'd expect this to be documented in.

Not sure about syntax for optional parameters, so I went with the usual brackets for optional items `[ ... ]`.

Refs #287

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -296,11 +296,11 @@ Passes two additional props to the base component: a state value, and a dispatch
 branch(
   test: (props: Object) => boolean,
   left: HigherOrderComponent,
-  right: HigherOrderComponent
+  [right: HigherOrderComponent]
 ): HigherOrderComponent
 ```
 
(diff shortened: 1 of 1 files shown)


istarkov:
Thank you BTW it looks untranslatable for me `the optional `right` higher-order component is applied or the wrapped component to the returned higher-order component`
May be because english is not my native language, BTW I can't get what means `or the wrapped component`

spoike:
Yeah, thinking about rewriting that. What I tried to write was that there seems to be two ways to use branch that are equivalent. IIRC it would be:

```
branch(props => !props.isLoading, <IsLoading />, <HasLoaded />);
```

and

```
branch(props => !props.isLoading, <IsLoading />)(<HasLoaded />);
```

The second use of branch returns a function (or higher order component) that takes a component to be wrapped (`<HasLoaded />`)

Would it make more sense to write it with the examples?

istarkov:
May be to write as is, as second param default is just an identity function

timkindberg:
This can be closed by #331

wuct:
Closed by #331

PR #327, opened 2017-03-07
Update branch related docs

timkindberg:
- Add example for renderNothing
- improve example for renderComponent
- Clarify optional param for branch

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -141,7 +141,7 @@ Takes an object map of handler creators or a factory function. These are higher-
 
 This allows the handler to access the current props via closure, without needing to change its signature.
 
-Handlers are passed to the base component as immutable props, whose identities are preserved across renders. This avoids a common pitfall where functional components create handlers inside the body of the function, which results in a new handler on every render and breaks downstream `shouldComponentUpdate()` optimizations that rely on prop equality.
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -317,16 +317,16 @@ This is useful in combination with another helper that expects a higher-order co
...
-    props => !hasLoaded(props),
+    isLoading,
     renderComponent(Spinner) // `Spinner` is a React component
   )
 
 // Now use the `spinnerWhileLoading()` helper to add a loading spinner to any
 // base component
 const enhance = spinnerWhileLoading(
-  props => props.title && props.author && props.content
+  props => !props.title && !props.author && !props.content
```
istarkov: `||` instead of `&&`
in your case will return `false` if one of fields will be defined

Also IMO better to not overuse !
and write `!(a && b)` instead of `!a || !b`

docs/API.md
```diff
@@ -343,7 +343,32 @@ const Post = enhance(({ title, author, content }) =>
...
+const hideIfNoData = hasNoData =>
+  branch(
+    hasNoData,
+    renderNothing
+  )
+
+// Now use the `spinnerWhileLoading()` helper to add a loading spinner to any
+// base component
+const enhance = hideIfNoData(
+  props => !props.title && !props.author && !props.content
```
istarkov: `||`

istarkov:
Super, thank you. Seems like need a fix.

timkindberg:
ok stuff is fixed per your feedback

timkindberg:
Moved to #331

PR #330, opened 2017-03-14
Add deep key evaluation to pick

timkindberg:


istarkov:
Do we expose pick in public API? 
I would prefer not to implement or expose deep pick, because its slower, has strange semantic (how property will be extracted), not the same as well known pick from lodash-underscore.

timkindberg:
Yeah I saw your other conversation in #300 but I'd already done the work for this. I thought maybe the duplicated efforts here would give evidence of a warranted feature. I'd actually wanted this for a training video I'm making. I'd wanted to do:

```
onlyUpdateForKeys(['chore.name', 'chore.complete'])
```

because my `chore` object has various fields that I *don't* want to update on. So my two cents is, I think it's a helpful feature, but I would not be hurt if it was left out ;)

PR #332, opened 2017-03-17
Add observableConfig for flyd streams

vic:


wuct:
Thanks! I'm not familiar with flyd, but over all looks good to me.

PR #333, opened 2017-03-17
Pass handlers object (self) to createHandler function

chrisui:
This allows creation and reference of handlers in a single `withHandlers` call rather than many.

```
withHandlers({
  a: (props, handlers) => (...args) => handlers.b(...args),
  b: () => (...args) => doStuff(...arg),
})
```

Posting this here as a spitball idea. Will wrap up with tests if we think this is a fair addition to recompose!

This avoids the following requirement.

```
withHandlers({
  b: () => (...args) => doStuff(...arg),
})
withHandlers({
  a: (props) => (...args) => props.b(...args),
})
```

Note: this has definitely popped up as a somewhat edge usage :thinking:

### DIFF
--- a/src/packages/recompose/withHandlers.js
+++ b/src/packages/recompose/withHandlers.js
@@ -27,7 +27,7 @@ const withHandlers = handlers => BaseComponent => {
           return cachedHandler(...args)
         }
 
-        const handler = createHandler(this.props)
+        const handler = createHandler(this.props, handlers)
         this.cachedHandlers[handlerName] = handler
 
         if (


istarkov:
Idea super! 
But have you tested your code as it seems like examples you provide will not work.
As `handlers` you use at `createHandler` are not the transformed handlers,
so to call `b` you need to call it like

```javascript
withHandlers({
  a: (props, handlers) => (...args) => handlers.b(props, handlers)(...args),
  b: () => (...args) => doStuff(...arg),
})
```

and this is not cool.

chrisui:
Totally not tested the code! Just did a quick psuedo diff to demo the idea :)

istarkov:
As I said idea super!!! ;-)

istarkov:
see https://github.com/acdlite/recompose/pull/401

Issue #337, opened 2017-04-04
Update documentation to include examples.


threehams:
There's an overall ticket for documentation with some larger goals, but I think that one or two examples per component / function can go a long way for the least amount of work.

Creating this issue for myself so I know where to focus, but I can update it as other pull requests are merged in.

I also included checklists on working with type annotations. I'm still figuring out how to use recompose with either Flow or TypeScript, so help would be appreciated there.

Higher-order components
- [x] mapProps()
- [ ] withProps()
- [ ] withPropsOnChange()
- [x] withHandlers()
- [ ] defaultProps()
- [ ] renameProp()
- [ ] renameProps()
- [x] flattenProp()
- [x] withState()
- [ ] withReducer()
- [ ] branch()
- [x] renderComponent()
- [x] renderNothing()
- [ ] shouldUpdate()
- [ ] pure()
- [x] onlyUpdateForKeys()
- [x] onlyUpdateForPropTypes()
- [ ] withContext()
- [ ] getContext()
- [ ] lifecycle()
- [ ] toClass()

Static property helpers
- [ ] setStatic()
- [ ] setPropTypes()
- [ ] setDisplayName()

Utilities
- [ ] compose()
- [ ] getDisplayName()
- [ ] wrapDisplayName()
- [ ] shallowEqual()
- [ ] isClassComponent()
- [ ] createEagerElement()
- [ ] createEagerFactory()
- [ ] createSink()
- [x] componentFromProp()
- [x] nest()
- [ ] hoistStatics()

Observable utilities
- [x] componentFromStream()
- [x] mapPropsStream()
- [ ] createEventHandler()
- [x] setObservableConfig()

TypeScript
- [ ] usage and limitations
- [ ] compose()
- [ ] type hinting with generics

TypeScript
- [ ] usage and limitations
- [ ] compose()
- [ ] type hinting with generics

istarkov:
For flow typings more information [here](https://github.com/acdlite/recompose/pull/241)

timkindberg:
@threehams I'm have a full length (~45min) course being released this month all about Recompose and Higher Order Components. I had to generate various examples in order to teach the concepts. Please feel free to link to, get inspiration from, migrate to plunkr, or steal my examples.

They are here: https://github.com/timkindberg/egghead-recompose

The folders are named the same as the lessons in my course.

[...]

PR #338, opened 2017-04-04
[WIP] Add examples to docs (#337)

threehams:
WIP, adding simple and real-world examples to the API documentation page. Since we're starting to use recompose more heavily at work, I'll cover details and examples for anything we're currently using.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -89,6 +89,41 @@ mapProps(
 
 Accepts a function that maps owner props to a new collection of props that are passed to the base component.
 
+This allows props to be derived from other props in a similar way to [reselect](https://github.com/reactjs/reselect).
+```js
+const User = ({ name, status }) =>
+  <div className="User">{ name }—{ status }</div>
+
+const UserList = ({ users, status }) =>
+  <div className="UserList">
+    <h3>{ status } users</h3>
+    { users.map((user) => <User {...user} />) }
+  </div>
+
+const users = [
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -89,6 +89,41 @@ mapProps(
 
 Accepts a function that maps owner props to a new collection of props that are passed to the base component.
 
+This allows props to be derived from other props in a similar way to [reselect](https://github.com/reactjs/reselect).
+```js
+const User = ({ name, status }) =>
```
istarkov: IMO examples which are not fit to screen is unreadable ;-)
mapProps is so simple, is it needed so big example for it?

threehams: `mapProps` is simple for us. For someone new to the library, new to FP in general, and unable to read the type signature (extremely confusing without any experience writing TypeScript/Flow definition files), it's not straightforward at all. You can see that by looking at comments in issues since last year. :)

I'm sure I can cut down the example to 1/2 or 2/3 the size, though.

It would definitely be easier to organize this if there were a dedicated site rather than a markup file, but this is more of an intermediate solution.

timkindberg:
Looking good!

threehams:
I'd like to cover lifecycle, withContext, and getContext, them that'll be enough for a first pass. Those three components caused me the most trouble when using the library for the first time.

istarkov:
We decided to add examples into project wiki

PR #340, opened 2017-04-06
Update recompose-relay to use relay-modern

chollier:
Relay2 aka Relay Modern is coming. With this new version we need (for now) to import createContainer from `react-relay/classic` (https://github.com/facebook/relay/releases/tag/v1.0.0-alpha.1)


chollier:
I also bumped the version to 1.0.0 as this would be a breaking change but I'm not sure it would be a MAJOR or MINOR

istarkov:
Thank you, please revert any changes in Version file, it will be changed automatically on npm deploy

chollier:
@istarkov just did so

istarkov:
Ooops, we have a problem - new react ;-)
image

istarkov:
I'll fix this in the other PR, but seems we have an issue with lifecycle component.

src/packages/recompose-relay/package.json
```diff
@@ -18,8 +18,8 @@
     "lodash": "^4.0.0"
   },
   "peerDependencies": {
-    "recompose": "^0.17.0 || ^0.18.0 || ^0.19.0 || ^0.20.0",
+    "recompose": "^0.17.0 || ^0.18.0 || ^0.19.0 || ^0.20.0 || ^0.21.0 || ^0.22.0",
     "react": "^0.14.0 || ^15.0.0",
-    "react-relay": "^0.6.0 || ^0.7.0 || ^0.8.0 || ^0.9.0"
+    "react-relay": "^1.0.0-alpha1"
```
istarkov: mb it will be better to also use `or` ?

chollier: I don't think so, granted that we change the import here, this new version of `recompose-relay` wouldn't be compatible with the previous versions of relay (there was no `react-relay/classic` export)

istarkov:
I don't know what to do, having that existing recompose users can? use stable relay versions. 
This change will break their ability to use it. @wuct @chollier any ideas?

chollier:
maybe release a recompose-relay@dev or @next like they on react-relay

wuct:
IMO it's fine to release at `v1.0` with a breaking change warning in the release note, and support old versions of relay at `v0.x`.

Secretmapper:
Bump?

idris:
@chollier does it make sense to have a "classic" function and a "modern" function?  Classic would use the `react-relay/classic` `createContainer`, whereas modern would use `createFragmentContainer`.

graingert:
Can we just pass relay into createContainer?

stevehollaar:
Since Relay modern has 3 container types (fragment, pagination, and refetch) with slightly different APIs, it may make sense to split this into 3 separate functions. 
Also I believe the `toClass` function can be omitted with Relay Modern.

Here is what I'm using currently:
```js
// @flow
import {
  createFragmentContainer,
  createPaginationContainer,
  createRefetchContainer,
  type GraphQLTaggedNode,
  type GeneratedNodeMap,
  type ConnectionConfig,
} from 'react-relay';

export const fragmentContainer = (fragmentQuery: GraphQLTaggedNode | GeneratedNodeMap) => (
  BaseComponent: React$Component<*>,
) => createFragmentContainer(BaseComponent, fragmentQuery);

export const paginationContainer = (
  fragmentQuery: GraphQLTaggedNode | GeneratedNodeMap,
  options: ConnectionConfig,
) => (BaseComponent: React$Component<*>) =>
  createPaginationContainer(BaseComponent, fragmentQuery, options);

export const refetchContainer = (
  fragmentQuery: GraphQLTaggedNode | GeneratedNodeMap,
  query: GraphQLTaggedNode,
) => (BaseComponent: React$Component<*>) =>
  createRefetchContainer(BaseComponent, fragmentQuery, query);
```

Example usage:
```js
compose(
  fragmentContainer(graphql`
    fragment Foo_viewer on User {
      id
    }
  `),
  ...
)(({ viewer }) =>
  ...
))
```

graingert:
I'd just like to get the peerDep updated as ASAP as possible, I think https://github.com/acdlite/recompose/pull/459 is the best bet for that.

PR #346, opened 2017-04-07
Fix lifecycle to support React 15.5

istarkov:
The reason: https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.createclass

Removing `createClass` at all from `lifecycle`.
(importing it from `create-react-class` will cause current package incompatibility with React 14, and all that `require` with version check is not better than current solution )

Also IMO `lifecycle` must be deprecated or even removed from recompose.

And yes possibly breaking change.

### DIFF
--- a/src/packages/recompose/lifecycle.js
+++ b/src/packages/recompose/lifecycle.js
@@ -1,4 +1,4 @@
-import { createClass } from 'react'
+import { Component } from 'react'
 import createHelper from './createHelper'
 import createEagerFactory from './createEagerFactory'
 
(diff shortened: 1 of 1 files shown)


TrySound:
Maybe just object assign? It is still exists in helpers.

istarkov:
You are right

istarkov:
@TrySound fixed, thank you!

pebie:
👍

romseguy:
> Also IMO lifecycle must be deprecated or even removed from recompose.

Why ? I'm wondering whether I should start using it in my project, thank you for your answer

wuct:
@romseguy you can get the idea from https://github.com/acdlite/recompose/issues/521#issuecomment-335309163.

PR #349, opened 2017-04-10
Reduce babel config

TrySound:


istarkov:
Will break common build as now all common modules contains import-export

istarkov:
Super now we are again have one source of babel plugins. Thank you.

PR #351, opened 2017-04-11
Add commonjs bundle as main entry point

TrySound:


istarkov:
So now just modules like `recompose/blablaMod` are now builded directly with babel, all other builds are using rollup.

istarkov:
Thank you!

PR #353, opened 2017-04-11
Prefer own shallowEqual

TrySound:


istarkov:
Thank you.
This method is differs from what is used for shallow equality at facebook (and at React)
https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/core/shallowEqual.js

And IMO install time dependency size does not matter. It's just few seconds if `fbjs` is not cached, and near zero if is.

TrySound:
36 385 -> 34 976
13 154 -> 12 943

wuct:
The fbjs version of shallowEqual uses `Object.is` polyfill from [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is), which handles 3 special cases:

```js
Object.is(0, -0);            // false
Object.is(-0, -0);           // true
Object.is(NaN, 0/0);         // true
```

If we don't handle these special cases, this change will be a breaking change.

neoziro:
fbjs is included in React, dropping it will not reduce final build size. Except if recompose can be be used with other library like preact or inferno.

PR #354, opened 2017-04-11
Add shallowEqualForKeys utility

TrySound:


istarkov:
Thank you. Should be really faster as I see?

TrySound:
~ three times faster I guess :)

TrySound:
Breaks a bit of logic without checking key existance and without NaN equality checking.

Issue #308, opened 2017-01-28
withState alternatives


istarkov:
Very good looking idea here https://twitter.com/jordwalke/status/825305448542724099 which can be used instead or with `withState`.

[some reasons why current implementation can possibly be changed](https://github.com/acdlite/recompose/pull/241#issuecomment-269765711)

So creating an enhancer which will wrap `setState` with

```javascript
updater = memoize(
  (stateUpdater, callback) => 
    (...args) => 
      this.setState(
        (state, props) => stateUpdater({ state, props },  ...args),
        callback
     )
)
```

will give us ability to use `updater` as in twit example

```javascript
const onInputValueChange = 
  ({ props, state}, event: { target: { value: inputValue }}) => ({  inputValue })

const Component = ({ updater, inputValue }) => (
  <input value={inputValue} onChange={updater(onInputValueChange)} />
)
```

Having that `updater` is memoized it will give the same function reference in `onChange` handler on every rerender.

wuct:
I believe Elm also use the similar API, and it looks promising to me.

istarkov:
Just have read this https://github.com/facebook/react/issues/8854 and there is possibility that this will be done in React, see statefull functuonal components, so in partial cases like all enhancers are referentially transparent, this will work from scratch.

wuct:
I am really looking forward to stateful functional component, but it will not land in 16 😞

slightlytyler:
@istarkov are you suggesting a new HOC for the library? `withValue`?

istarkov:
@slightlytyler there are some problem with this HOC implementation, the same problem have guys with react bindings for reasonml. 
The problem that `updater` should do an atomic updates using react syntax `setState((state, props) => ...)` and currently in React there's no opportunity for you to say "don't update state if it's not needed" 
So every call to `updater someHandler`will cause a state refresh and rerender even if state will not be changed.
For now react bindings for reasonml uses non atomic updates. Which in some rare situations can cause a problems.

timkindberg:
Picking up the conversation from #321 (which would also solve for #243).

I love that you are looking ahead to future apis, and I think adding an alternative signature until then is not a bad thing. There's no reason to wait. Especially for such a simple, backwards compatible and non-controversial change. After reviewing this [comment](https://github.com/acdlite/recompose/pull/241#issuecomment-269765711) I've updated the proposal:

[...]

PR #357, opened 2017-04-12
Add class-like state behavior option to withState

### LINKED ISSUE #321: withState that has same API as React state
I found myself trying to use `withState` to recreate my own HOC that has the same `state` & `setState` API that a normal React class component has. The behavior (as we all probably know) is where `setState()` merges passed objects with existing state. The API is perfectly suited for managing local state.

I then wondered would it be more generally useful and also immediately familiar if `withState` mimicked the React state api?

Example:
```jsx
const initialState = { foo: true };
const enhance = withState(initialState); // only 1 param needed
const Example = enhance(({ state, setState }) => // always 'state' and 'setState'
  <div>
    <button onClick={ () => setState({ bar: true }) } />
    <div>{ state.foo } and { state.bar }</div>
  </div>
);
```
Notice how I arbitrarily added `bar`. This version is very flexible.

I do like the current `withState`, but Recompose is the building blocks for custom HOCs. The current `withState` would be more easily composed using the `withState` above other than vice versa. Like so:

```jsx
const legacyWithState = (name, method, initial) => compose(
  withState({ val: initial }),
  mapProps(({ state, setState }) => ({
    [name]: state.val,
    [method]: (val) => setState({ val })
  }))
);

```

Thoughts?

timkindberg:
Adds an additional optional param set for withState, where if you only provide initialState, then stateName and stateUpdaterName are defaulted to 'state' and 'setState'. In this form setState works exactly like setState from a typical class component (because its really just proxying straight to `this.setState`).

Example Usage:
```javascript
import { withState } from 'recompose';

// setting initial state only
const enhance = withState({ red: 0, green: 0, blue: 0 });

// automatically get state and setState props
const RGBChooser = enhance(({ state, setState }) => 
  <div>
    <input type="text" value={ state.red } onChange={ (e) => setState({ red: e.target.value }) } />
    <input type="text" value={ state.green } onChange={ (e) => setState({ green: e.target.value }) } />
    <input type="text" value={ state.blue } onChange={ (e) => setState({ blue: e.target.value }) } /> 
    <div
      style={{
        display: block,
        width: 30,
        height: 30,
        backgroundColor: `rgb(${state.red},${state.green},${state.blue})`
      }}
    /> 
  </div>
);
```
I have code, tests, and docs in this PR.

closes #321

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -243,13 +243,14 @@ const Post = enhance(({ title, content, author }) =>
 
 ```js
 withState(
-  stateName: string,
-  stateUpdaterName: string,
+  stateName: ?string,
+  stateUpdaterName: ?string,
   initialState: any | (props: Object) => any
 ): HigherOrderComponent
 ```
+Passes two additional props to the base component: a state value, and a function to update that state value.
 
-Passes two additional props to the base component: a state value, and a function to update that state value. The state updater has the following signature:
+If stateName and stateUpdateName are provided then the behavior is as follows. The state updater has the following signature:
 
 ```js
 stateUpdater<T>((prevValue: T) => T, ?callback: Function): void
@@ -273,7 +274,20 @@ The second form accepts a single value, which is used as the new state.
 
 Both forms accept an optional second parameter, a callback function that will be executed once `setState()` is completed and the component is re-rendered.
 
-An initial state value is required. It can be either the state value itself, or a function that returns an initial state given the initial props.
+Only the initialState param is required. It can be either the state value itself, or a function that returns an initial state given the initial props.
+
+If stateName and stateUpdaterName are not provided then they default to `'state'` and `'setState'`. In this form, state and setState behave exactly like setState in a typical react class. So `state` will be an object and `setState()` will accept a new state object to merge with the existing state. In this form `setState()` also still allows either an object or function as a param for updating state:
+
+```js
+const addCounting = compose(
+  withState({ count: 0 }),
+  withHandlers({
+    increment: ({ setState }) => () => setState(({ count }) => count + 1),
+    decrement: ({ setState }) => () =>  setState(({ count }) => count - 1),
+    reset: ({ setState }) => () => setState({ count: 0 })
+  }))
+)
+```
 
 ### `withReducer()`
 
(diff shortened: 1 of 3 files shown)


istarkov:
I'll write here why I would prefer something like #308 and not a current solution.
Because current solution causes new function creation for every handler. Yes we can always use `withHandlers` but having that in withHandlers method call like

```javascript
withState('something', 'setSomething', {a:1, b: 2}),
withHandlers({
  onAChange: ({ setSomething }) => ({ a }) => ({
      setSomething(prev => ({...prev, a}))
  })
})
```

As you see this is not so a big addition even you will write it inside render.
So I don't think solution above will somehow make better current situation.

I created an issue here https://github.com/reasonml/reason-react/issues/37 about what I wrote in #308 BTW we can try implementation provided there https://github.com/reasonml/reason-react/blob/master/src/reactRe.re#L422-L454 and even it has some drawbacks now 
I think it will be solved by React soon.

timkindberg:
I totally agree with you and are we sure it will be solved by React soon? Will they even settle on a final API soon as well? If not, why wait? Let's improve upon what we know now.

True, this is not a clever feature, but this is probably how withState should have behaved originally.

Typically I agree that most PRs should be rejected because they can be implemented in dev's own code, but this is simply allowing functional stateless components to have state features that mirrors existing class state features. For beginners of recompose it will be very comfortable, and it's way more flexible for when devs need to manage more than a single state variable.

istarkov:
> but this is probably how withState should have behaved originally

agree, btw original React setState with state merging is also confusing, year or more ago I did
a PR to react-redux https://github.com/reactjs/react-redux/pull/50 so even superexperienced developer did a mistake with original setState

istarkov:
I wanna say current behavior is more predictable, and in a lot of cases allows to use it without withHandlers  etc

timkindberg:
> lot of cases allows to use it without withHandlers etc

When can you ever use it without withHandlers? Or do you mean use an anonymous function with the jsx?

> So until we get additional setState version in React, all will be as is.

Can't we fix it ourselves instead of waiting on react? Can we throw and error to escape setState and have a wrapper/proxy setState that catches the error? In this way we could implement what you want with the updater approach.

PR #365, opened 2017-04-19
Update eslint, add prettier (with lint-staged and husky)

istarkov:


package.json
```diff
@@ -12,7 +12,9 @@
     "test": "cross-env BABEL_ENV=cjs jest --coverage",
     "test:watch": "npm test -- --watch",
     "release": "node scripts/release.js",
-    "postinstall": "node scripts/installNestedPackageDeps.js"
+    "postinstall": "node scripts/installNestedPackageDeps.js",
+    "format": "rm -rf src/packages/*/node_modules;prettier --semi false --trailing-comma es5 --single-quote --write '{src/,scripts}/**/*.js'",
```
evenchange4: It is a little inconvenient if `node_modules` have been removed per commit, we need to re-install it while developing later. How about adjusting pattern something like:

```json
"format": "prettier --semi false --trailing-comma es5 --single-quote --write 'scripts/*.js' 'src/packages/*/*.js' 'src/packages/*/!(node_modules)/**/*.js'",

```

istarkov: I will try, thank you. (I havent found a pattern :-))

istarkov: Fixed, can't get why it works but it works, I thought multiple path templates are combined using OR operation but here is something different. May you say where I can read about?

.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 release
 lib
 coverage
+.vscode
```
MichaelDeBoey: This should be in your global `.gitignore` file, because this is specific to your own machine.

istarkov: Yep, my fault.

PR #370, opened 2017-04-30
Treeshake

TrySound:
Tried to treeshake whatever possible. But shallowEqual still fails because of commonjs. Bad dependency.


istarkov:
I see no issues, @wuct please merge if all is ok from your side.

istarkov:
We need to think about moving fbjs `shallowEqual` into the recompose, this will solve that `shallowEqual` is not tree shaked and included with every import.

TrySound:
Or better, I introduce es modules in fbjs to make happy everybody.

wuct:
Should we open a new issue for shouldComponentUpdate?

TrySound:
In react? Or you are about shouldUpdate hoc?

wuct:
Oh, I mean recompose/shallowEqual. I am typing on my phone. sorry for the mistake.

wuct:
I am going to merge this PR. Let's track shallowEqual in other issue.

TrySound:
ShallowEqual export is tree shaked. Fbjs/shallowEqual dep is not.

TrySound:
Okay

wuct:
@TrySound Thanks for your great work!

PR #373, opened 2017-05-04
Export `componentFromStreamWithConfig()` and `mapPropsStreamWithConfig()`

mrapogee:
I've added tests + exported mapPropsStreamWithConfig and componentFromStreamWithConfig.

How would you recommend changing the docs? I'm thinking we should recommend users but something like this:

```
export const componentFromStream = componentFromStreamWithConfig(rxjsConfig)
```

from a project utils directory. Do you have a pattern for depreciating setObservableConfig?


.gitignore
```diff
@@ -3,3 +3,4 @@ release
 lib
 coverage
 .vscode
+yarn.lock
```
istarkov: please remove this line, yarn.lock must be included included

docs/API.md
```diff
@@ -728,6 +728,18 @@ const Counter = componentFromStream(props$ => {
 })
 ```
 
+### `mapPropsStreamWithConfig()`
```
istarkov: description, example?

package.json
```diff
@@ -78,7 +78,7 @@
     "rxjs": "^5.0.0-beta.9",
     "shelljs": "^0.6.0",
     "sinon": "^1.17.1",
-    "webpack": "^2.4.1",
+    "webpack": "^2.5.0",
```
istarkov: No need to upgrade webpack for this commit.

src/packages/recompose/__tests__/componentFromStream-test.js
```diff
@@ -1,11 +1,28 @@
...
+import { Stream as MostStream } from 'most'
 import rxjsConfig from '../rxjsObservableConfig'
-import componentFromStream from '../componentFromStream'
+import mostConfig from '../mostObservableConfig'
+import { componentFromStreamWithConfig } from '../componentFromStream'
 
-setObservableConfig(rxjsConfig)
+const componentFromStream = componentFromStreamWithConfig(rxjsConfig)
+
+test('componentFromStreamWithConfig creates a stream with the correct stream type.', () => {
```
istarkov: Isn't it will be better to create `componentFromStreamWithConfig-test.js`?

src/packages/recompose/componentFromStream.js
```diff
@@ -3,8 +3,18 @@ import { createChangeEmitter } from 'change-emitter'
 import $$observable from 'symbol-observable'
 import { config as globalConfig } from './setObservableConfig'
 
-export const componentFromStreamWithConfig = config => propsToVdom =>
-  class ComponentFromStream extends Component {
+const identity = v => v
+
+export const createObservableConfig = rawConfig => ({
```
istarkov: Why? What the reason to change current implementation? No need to merge the config as in recompose implementation `toESObservable`,`fromESObservable`  are defined.
It's not the `options` object with tens of props.

istarkov:
Thank you, see the review comments, and also please use `prettier` for code formatting.

docs/API.md
mrapogee: This is WIP, I had a couple of questions about whether you wanted to depreciate setObservableConfig.

.gitignore
mrapogee: Sounds good.

src/packages/recompose/componentFromStream.js
mrapogee: The tests fail if this isn't here, toESObservable is undefined for a couple of the observable configs. 

Should I instead define  toESObservable in all observable configs?

istarkov: I think yes, it will be better, more strict `config` less implicit magic.

docs/API.md
istarkov: `setObservableConfig` is not so bad even it uses globals, in our daily work we use a lot of implicit configs at compile level - all that webpack, babel etc, so I'm not sure about deprecation.

mrapogee: It's really bad for libraries though. We should at least note not to use it if you're using it to create a library.

istarkov: I know, now we will have an alternative solution for libraries, but I see no bad to provide defaults for an apps in an easy way.

istarkov:
It will be better to revert `yarn.lock` and `package.json`

mrapogee:
There is no current yarn.lock in master, but I reverted the package.json

istarkov:
:-) recompose does not use yarn, npm for now ;-) so just delete it ;-)

mrapogee:
Sounds good, can I just .gitignore it again incase someone else runs yarn instead of npm?

istarkov:
better just delete and then move the whole project on yarn + lerna I think

istarkov:
also still have prettier lint errors, `npm run lint`

mrapogee:
Okay, I fixed the lint errors.

istarkov:
Seems like because of prettier changed fast we got a new lint errors, seems like to avoid this we need to switch on yarn asap.
BTW travis still contains errors, I think because of yarn and non npm is used in your local env.
I recommend to `rm -rf node_modules;npm install;npm run format;npm run lint;`
sorry for this, we will fix that in near future.

mrapogee:
Okay, looks like that did it!

istarkov:
Thank you, please check 0.23.2

PR #374, opened 2017-05-08
Change createClass to Component in docs

Klaasvaak:
With `v0.23.0` `createClass` was replaced with `Component` but some docs where not updated.


istarkov:
Thank you!!!

PR #391, opened 2017-05-26
Fixed lack of es6 Object.assign in target envs like IE11

Andarist:
Fixes 'regression' introduced in - https://github.com/acdlite/recompose/pull/346 , which also should result in some minor perf/memory gain as `spec` gets shared on the prototype now and this needs to be done only once.

cc @istarkov

### DIFF
--- a/src/packages/recompose/lifecycle.js
+++ b/src/packages/recompose/lifecycle.js
@@ -15,12 +15,6 @@ const lifecycle = spec => BaseComponent => {
   }
 
   class Lifecycle extends Component {
-    constructor(...args) {
-      super(...args)
-
-      Object.assign(this, spec)
-    }
-
     render() {
       return factory({
         ...this.props,
@@ -28,6 +22,8 @@ const lifecycle = spec => BaseComponent => {
       })
     }
   }
+
+  Object.keys(spec).forEach(hook => (Lifecycle.prototype[hook] = spec[hook]))
 
   if (process.env.NODE_ENV !== 'production') {
     return setDisplayName(wrapDisplayName(BaseComponent, 'lifecycle'))(


istarkov:
Seems like we can return back previous version, as now it's known that `create-react-class` supports older react versions.

istarkov:
(known - means I read it somewhere in twitter ;-))

Andarist:
Not sure we should return back to the `create-react-class` - its just another dependency, which is not that useful. Note in the docs about which package range support which React version should be quite enough.

Other thing is what browsers do we want to support out of the box (without need of polyfills to be provided etc)

pebie:
👍

istarkov:
There are linting errors in this PR, I can't accept it until it will be fixed.

Andarist:
@istarkov fixed, had somehow misconfigured my editor and it didnt highlight the errors in the project

istarkov:
There were a bug in previous fix I did, 
React.createClass has `autobinding` behaviour for all methods except `lifecycle` and methods in argument object prototype. 

Don't you think we need this here?

Andarist:
The methods are put on the prototype so the `this` should be referenced correctly once the components get instantiated

istarkov:
How you will explain this if proto members referenced correctly?
```
class MClass {}; 
MClass.prototype.hello = function() { console.log(this);}; 
var a = new MClass(); 
a.hello(); // output undefined
```

istarkov:
Ooups I was wrong sorry ;-)

istarkov:
Thank you!!!

istarkov:
Released
https://github.com/acdlite/recompose/releases/tag/v0.23.5

PR #396, opened 2017-06-01
Fix missing parenthesis

rtivital:


TrySound:
It's not missing. Don't forget we wrap jsx with function call

rtivital:
Oh, sorry for that then, looks very confusing without parentheses around multiline JSX

Issue #368, opened 2017-04-28
How to change state from outside component


dingchaoyan1983:
now I am using recompose to create HOC for my stateless component, and I use `withState` to hold the inner state in the HOC, and this state is initial from the pass in props. this is work fine in the mount period, but I don't know how to handle the state change from props at the `update` period, i just want to use lifecycle `componentWillReceiveProps` to handle that, but nothing happened. so anyone can help me on this?

the code just like below:
```
compose(
    withState('counter', 'updateCounter', ({ outerCounter }) =>outerCounter ),
    lifecycle({
       componentWillReceiveProps(nextProps) {
           if(this.props.outerCounter !== nextProps.outerCounter) {
                 this.props.updateCounter(nextProps.outerCounter);
           }
        }
     })
)({ counter, updateCounter } => (
<div>
    <div>{counter}</div>
    <button onClick={
        () => updateCounter(counter++)
    }>Add counter</button>
</div>
));
```
we can see my purpose is to pass a counter the component, and click the button we can change the counter, but we also can change the counter from outside of component, it can also effect the inner value. I don't know how to implement it by using `recompose`. I know we can update state at  lifecycle method `componentWillReceiveProps`, but seems it hard to write it in recompose way

istarkov:
Hi,
At first I want to say that state dependent of props is not a good idea and always can be fixed by moving it upper in a tree. Sometimes can be fixed by using props and state to calculate anything you need on the fly without updating state itself.
Why it's not good idea: 
- Having such dependent state means that your app state can't be mapped directly to a view, as some of your component state depends not on data but on data change. (If it does not depend on data change you always can calculate it on the fly so you don't need to update it)
- Easy to make a mistake as you need to track such state in a few places in a same manner - constructor, receiveProps etc

But, sometimes it's hard to move state upper in a tree, sometimes we need to encapsulate such logic inside component without affecting other components, and sometimes we just have no time, and even I wrote that it's not good idea, I can't say that it's very bad idea. 

In recompose the only good way to work with such state is to use `mapPropsStream` enhancer, as all other ways have drawbacks, so in simplest form it will be something like this

```javascript
mapPropsStream(props$ => props$.scan( /* here logic to update state */ ))
```

dingchaoyan1983:
yes I agree with you, but sometime we need encapsulate the state with component that will make the component expose less api for other users( developers ), so this is why I need put the state into the component scope, right now i have no good idea to handle it.

evenchange4:
[...]

PR #397, opened 2017-06-03
Create withPropState.js

garretttaco:
One of the use cases of this is to allow a controlled input to be stored locally in state with an initial value from props. The initial value will most likely be coming from a global store. The user can then have an onBlur of the input field to save the value. If the store is updated via websockets or short/long pulling and the value is saved and stored in the store then this component will reflect that change by updating the local state of `withPropState` in the `componentWillReceiveProps` lifecycle method. This component works just like `withState`, except it updates the state whenever the prop passed in.

Example usage:

```
compose(
  ...
  withPropState('propValue', 'updatePropValue', ({ propValue }) => propValue),
  ...
)
```

### DIFF
--- a/src/packages/recompose/withPropState.js
+++ b/src/packages/recompose/withPropState.js
@@ -1,0 +1,43 @@
+import { Component } from 'react'
+import setDisplayName from './setDisplayName'
+import wrapDisplayName from './wrapDisplayName'
+import createEagerFactory from './createEagerFactory'
+
+const withPropState = (stateName, stateUpdaterName, initialState) => {
+  return BaseComponent => {
+    const factory = createEagerFactory(BaseComponent)
+    class WithPropState extends Component {
+      state = {
+        stateValue: typeof initialState === 'function' ? initialState(this.props) : initialState,
+      }
+
+      componentWillReceiveProps(nextProps) {
+        this.setState({ [stateName]: nextProps[stateName] })
+      }
+
+      updateStateValue = (updateFn, callback) =>
+        this.setState(
+          ({ stateValue }) => ({
+            stateValue: typeof updateFn === 'function' ? updateFn(stateValue) : updateFn,
+          }),
+          callback,
+        )
+
+      render() {
+        return factory({
+          ...this.props,
+          [stateName]: this.state.stateValue,
+          [stateUpdaterName]: this.updateStateValue,
+        })
+      }
+    }
+
+    if (process.env.NODE_ENV !== 'production') {
+      return setDisplayName(wrapDisplayName(BaseComponent, 'withPropState'))(WithPropState)
+    }
+
+    return WithPropState
+  }
+}
+
+export default withPropState


istarkov:
Its really bad idea when state of your component depends on render and not on data. We will never accept this. And I highly recommend you to not use this.

istarkov:
Want to add:
Also your enhancer will not solve even your case:
Let's think a little: 
1) User change input
2) User blur input 
3) Save to server begin
4) User focus input and change input again
5) Response from server received and overwrites what user has entered at step 4

As you could see it's not what user expect.

In all similar cases all you need is to move local state at upper level, for example in redux reducer etc... In your case you will need an additional flag like `dirty` saying that user changing the input now, to prevent `5` above.

istarkov:
Want to add: there are libraries for optimistic updates available like https://github.com/mattkrick/redux-optimistic-ui they allow you to use one global state and solves case above in other very interesting way.

istarkov:
Also I wrote about similar idea here https://github.com/acdlite/recompose/issues/368#issuecomment-297917744

garretttaco:
Thanks for the feedback. I definitely feel like this is a problem to solve. Maybe my use case above was not the best example. I will have to think this over. Upon reading your comment from #368, I didn't gain that much more insight into what to do instead. Do you have maybe a better example of how to accomplish this? I feel like optimistic updating is a different issue. I get that it will solve the above issue that I presented with the controlled input and the data being saved to the server. However, what happens when I have data that will be broadcasted via websockets? How will I get that to reflect into my component? Will I just need to track both the controlled input state (be it component state or via redux-optimistic-ui) and the server values? Then intelligently merge them each time one changes?

Thanks again for the feedback and knowledge share, I truly am trying to learn the best practice for this problem. 👍

istarkov:
About redux optimistic libs, as I remember (it was long time ago I played with some of them) the idea was:
Every `save` action internally creates an action something like `{type: 'BEGIN_SAVE', payload}` where payload is an optimistic result of action. Before dispatching this action, optimistic library saves a snapshot of current state. Then action dispatched and all reducers can immediately accept that optimistic result without waiting a server response.

Then server can respond with two action types:  `COMMIT` or `ROLLBACK`, if action type is `COMMIT` - all is fine you don't need to do anything, you are already have a good state. 
If action is `ROLLBACK` the library internally get a saved snapshot of the state, and then replays all actions dispatched after `BEGIN_SAVE` so at the end you have a state like that save action was not occurred. (You can show error message that result wasn't accepted bla bla) 
So you don't need to merge anything using optimistic libraries. 

Now about  
> I didn't gain that much more insight into what to do instead

The general idea is to think about your component as about some kind of pure function like `view = fn(state)`
Every side effect inside - like changing the internal state based on outer state or on render can be a problem, as function will not be pure, and your view will be unpredictable. 
How to fix this - if you move internal state into the outer state (and all the changing logic out of view), nothing will be changed inside your function  - all is fine - so the general solution is to move state out of your component at upper level.

BTW in real life as I wrote sometimes we can't do that in such case I recommend to use `mapPropsStream`  as `rx` allows easily to combine internal state with outer, or to use naked react classes without trying to implement such logic using `recompose`

wuct:
Single source of truth is always the best practice.

PR #398, opened 2017-06-07
Create ifElse utility helper

JAFong:
ifElse()

ifElse(
  test: (props: Object) => boolean,
  left:  (props: Object) => any,
  right:  ?(props: Object) => any
): HigherOrderComponent

Accepts a test function and two functions. The test function is passed the props from the owner. If it returns true, the left function will be invoked; otherwise, the right function will be invoked if it is supplied.


istarkov:
Thank you, but 
we are already have branch enhancer with same functionality

istarkov:
Also please take a look into your implementation

```javascript 
if (test(props)) {
  left(props)
} else if (right) {
   right(props)
}
```
results of `left` `right` are not used  anywhere. So we can think about `left` `right` as about functions without return `left: (props: Object) => void,`

Any `void` function or useless or has side-effects. So it's not a good idea to use such enhancer at all.

Please take a look into `branch` implementation to get the idea how it works.

ha404:
🤔  Hmm, what if we want this to not return a component but just to run some simple functions with no side effects:
```js
ifElse(
  props => props.isUser,
  props => props.history.push('/'),
)

ifElse(
  props => props.isFeatureEnabled,
  props => props.track('Test A'),
  props => props.track('Test B'),
)
```

istarkov:
@ha404 In such case just use `withProps` and js language
`withProps(props => test(props) ? trackA : trackB)`

BTW I recommend to not use history change on prop change it will be better to move out side effects out of your Component

ha404:
I see, how would you deal with a history change outside of a Component? Isn't everything a component?

istarkov:
Im not a Component :-) so everything is not a component.
I meant move side-effects out of rendering pipeline,  without additional libs you are free to use them at event handlers - (in your case it was some event which caused change of isUser prop, so history push can be placed directly there)
With libs like redux I prefer to dispatch plain  actions in handlers and use redux-observable or redux-saga for processing and side-effects (even with libs I never dispatch anything at rendering stage, its always a reaction on user-network-timer etc events, not a reaction on properties change)

ha404:
Why can't all our components get along? 

I still don't understand how that will help authenticate my components. Like, what if I wanted to protect an `Admin` component on an `/admin` route? How would observing the event that a user authenticated help me block that user from viewing the `Admin` component on my `/admin` route?

istarkov:
I haven't wrote that you can't use Components alone. I wrote that you need to avoid side-effects in rendering pipeline. 
The reasons are:
1) It's easier to work when you components have no side-effects and just a functions of data (_all that nice words like testability, predictability, support etc what it gives you from functional programming here_)
2) Some side-effects like history.push can cause immediate state change in above components wrappers like in react-router, React has a strong restriction from what lifecycle methods you can call state change - easy to make a mistake
3) Having that component has already started to render and you call history.push inside right lifecycle handle you will got the situation that `history.push` will not break current rendering process it's just updates state at some component at upper level so you need an additional logic to prevent showing current view.

All problems are easy to avoid intercepting route transition events and preventing route change in situations you need. This also make your Components more dumb, as all the knowledge about routing, auth, histories will be implemented out of Views.

BTW you can write all that logic inside Components as nor js nor React force you to write pure side-effects less Components.

ha404:
I agree, I understand the benefits of avoiding the side effects. I'm just having a hard time understanding literally where should I put my authentication guard in the code? Would that be somewhere in React Router? Do I do something like [this](https://reacttraining.com/react-router/web/example/auth-workflow) vs. use a HOC?

ha404:
JK, I talked myself out of this. Redirection should never be handled by individual components.

PR #399, opened 2017-06-07
Run eslint on staged files as part of the precommit hook

Andarist:
It will guard contributors before commiting with lint errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   "lint-staged": {
     "*.js": [
       "prettier --semi false --trailing-comma es5 --single-quote --write",
+      "eslint --fix",
       "git add"
     ]
   },


istarkov:
https://github.com/acdlite/recompose/blob/master/package.json#L11

istarkov:
Also seems like time to move on yarn ;-) based on travis errors

Andarist:
Yeah, ive seen the lint task - but its quite different. Running `eslint` is only reporting errors, but running it with `--fix` may auto-correct some simple mistakes. Thought that would be more suitable as the part of the `precommit`.

Let me know what do u prefer, I'll amend the commit as u wish. Aint sure though if moving to yarn now is so desirable, when npm5 is already out, but thats also what im thinking myself - dont have yet strong opinion, time will tell I guess.

istarkov:
I meant the folders for linter, must be the same as in `lint` task.

As I know `npm 5` have a lot of issues for now, without `yarn` all that new `prettier` plugin errors will be very often.

Andarist:
as to the `lint-staged` - only staged files are geting passed to its `*.js` rule, so as only those files which are really geting added to the repo (no build directories, node_modules etc) that basically already covers ur `scripts src` scope, should i changed it to `npm run lint -- --fix` regardless?

hm, honestly I dont know what (i might be not aware of smth) how yarn will help with eslint's prettier plugin, could u elaborate? or is just connected to the versioning of the prettier plugin which introduces breaking changes rly often? If yes then it should be sufficient to add `package-lock.json`, right?

from what I see `scripts` are breaking some rules from the prettier plugin, gonna try to fix it

istarkov:
Got about lint-staged, you are right, all is fine.
yarn with lock file will lock a plugin version.
To fix rules we have `format` script in package.json.

istarkov:
Thank you!!!

Andarist:
> yarn with lock file will lock a plugin version.

same thing should happen when using `package-lock.json` when using npm5, right?

istarkov:
Yes

PR #401, opened 2017-06-07
Allow to call internal handlers within withHandlers

istarkov:
I like the idea described here #333
BTW seems like author don't want to finish it.

The idea is to allow to call internal handlers within withHandlers, as for now it's not a rare case for me to create a chain of withHandlers just to allow reuse some handlers.

Usage example

```javascript
withHandlers({
    selfHandler: props => val => {
      selfHandlerCallSpy({ props, val })
    },
    handler: (_, { selfHandler }) => val => {
      selfHandler(val) // it works
    },
})
```

Before to do that I did:

```javascript
withHandlers({
    selfHandler: props => val => {
      selfHandlerCallSpy({ props, val })
    },
}),
withHandlers({
    handler: ({ selfHandler }) => val => {
      selfHandler(val) // it works
    },
})
```

BTW I'm not sure do we need this or not, as even I like the idea I'm pretty happy with withHandlers chain

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -129,10 +129,10 @@ Instead of an array of prop keys, the first parameter can also be a function tha
 ```js
 withHandlers(
   handlerCreators: {
-    [handlerName: string]: (props: Object) => Function
+    [handlerName: string]: (props: Object, self: Object) => Function
   } |
   handlerCreatorsFactory: (initialProps) => {
-    [handlerName: string]: (props: Object) => Function
+    [handlerName: string]: (props: Object, self: Object) => Function
   }
 ): HigherOrderComponent
 ```
@@ -143,18 +143,25 @@ This allows the handler to access the current props via closure, without needing
 
 Handlers are passed to the base component as immutable props, whose identities are preserved across renders. This avoids a common pitfall where functional components create handlers inside the body of the function, which results in a new handler on every render and breaks downstream `shouldComponentUpdate()` optimizations that rely on prop equality. This is the main reason to use `withHandlers` to create handlers instead of using `mapProps` or `withProps`, which will create new handlers every time when it get updated.
 
+self argument contains all handlers defined in current enhancer.
+
 Usage example:
 
 ```js
 const enhance = compose(
   withState('value', 'updateValue', ''),
   withHandlers({
-    onChange: props => event => {
-      props.updateValue(event.target.value)
+    onLog: props => message => {
+      console.log(message, props)
     },
-    onSubmit: props => event => {
+    onChange: (props, self) => event => {
+      props.updateValue(event.target.value)
+      self.onLog('onChange')
+    },
+    onSubmit: (props, self) => event => {
       event.preventDefault()
       submitForm(props.value)
+      self.onLog('onSubmit')
     }
   })
 )
@@ -167,6 +174,7 @@ const Form = enhance(({ value, onChange, onSubmit }) =>
   </form>
 )
 ```
+
 
 ### `defaultProps()`
 
--- a/src/packages/recompose/__tests__/withHandlers-test.js
+++ b/src/packages/recompose/__tests__/withHandlers-test.js
@@ -162,3 +162,37 @@ test('withHandlers allow handers to be a factory', () => {
   expect(componentHandlers[0]).toEqual(componentHandlers2[0])
   expect(componentHandlers[0]).not.toBe(componentHandlers2[0])
 })
+
+test('withHandlers allows to call internal handlers', () => {
+  const selfHandlerCallSpy = sinon.spy()
+
+  const enhance = withHandlers({
+    handler: (_, { selfHandler }) => val => {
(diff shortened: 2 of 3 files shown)


lucasfeliciano:
🎉 
Awesome!!

Gustavo0570:
+1

wuct:
Why not just

```js
withHandlers(() => {
  const selfHandler = props => val => {
    selfHandlerCallSpy({ props, val })
  }

  return {
    selfHandler,
    handler: (props) => val => {
      selfHandler(props, val) 
    },
  }
})
```

Goluis:
this can break code like this:
```js
compose(
  withState('email', 'changeEmail', ''),
  withHandlers({
    changeEmail: props => e => {
      props.changeEmail(e.target.value)
    },
    onSubmit: props => e => {
      e.preventDefault()
      
      props.changeEmail('')
      props.submit()
    }
  })
)(Component)
```

istarkov:
It cant, self handlers are passed not via props

Goluis:
in the example `changeEmail` and `onSubmit` are using the `changeEmail` function from `withState`, with this pr `onSubmit` will use the `changeEmail` inside `withHandlers` and it will not work.

istarkov:
It will not, as your code is wrong. self changeEmail is not passed via props

istarkov:
To be clear current PR does not change previous behaviour. It just add additional argument alongside with props

Goluis:
oh... if it's with an additional parameter the code will not broke :+1: 
in any case, the code work, the `changeEmail` that the handlers are using is the one from `withState`

lucasfeliciano:
In my opinion this should be transparent to be able to call other handlers from handlers, no need to pass selfHandler or self .

wuct:
We also need to update the type def.

istarkov:
Im not sure we need this :-) So not sure we need typedefs

wuct:
IMO we don't need this. Just assign the reused handler to a variable inside
the closure is good enough to me.
Ivan Starkov <notifications@github.com>於 2017年8月15日 週二，下午5:21寫道：

cbranch101:
Having to double invoke every wrapped handler like that seems unnecessarily annoying.  There's already a convention like this is many other library where they access `ownProps` in some way.  This really does seem like it would be very helpful

istarkov:
Imo it depends. We are always have some edge cases. Every simple library after some time becomes not so simple because of that. For now I think that its not the reason to change the simple implementation.
For now for me simplicity matters more than ability to support every user case.
I commercial world this isnt true, but this is opensource

rek:
To access a handler from inside another one, currently I do something like:

	```
withHandlers(() => {
		const self = {
			handlerX: (props) => (e) => {
				console.log(e)
			},
			handlerY: () => (e) => {
				self.handlerX(e)
			}
		}
		return self
	})
```

Was confusing to work this out the first time though, it would be nice if they were just in props.

**Comment by fmedinac** [comment]
@rek The problem with your code, if I'm not wrong, is that you are passing `e` as your `handlerX`´s `props`. The correct way would be:
```
withHandlers(() => {
		const self = {
			handlerX: (props) => (e) => {
				console.log(e)
			},
			handlerY: (props) => (e) => {
				self.handlerX(props)(e)
			}
		}
		return self
	})
```

The problem then would be that your `handlerX` would get a not-up-to-date `props` (it would consider the props sent - in the moment - to handlerY). Which would force you to do something like:

```
withHandlers({
  getProps: props => () => props,
}),
withHandlers(() => {
		const self = {
			handlerX: (props) => (e) => {
				console.log(e)
			},
			handlerY: (props) => (e) => {
				self.handlerX(props.getProps())(e)
			}
		}
		return self
	})
```

**Comment by Billy-** [comment]
I would love this feature. +1

**Comment by vampolo** [comment]
I would love to have this feature too.

Modern Javascript is using arrow functions instead of closures. When writing a React component, without recompose, you expect methods to be able to reference each other.
If `withHandlers` is meant to replace the methods you would add in a React component, then it seems natural to add the support. Or maybe `withHandlers` is not meant to be so?

**Comment by wuct** [comment]
> Modern Javascript is using arrow functions instead of closures. 
@vampolo Recompose does use arrow functions and closures a lot internally. Recompose is more about functional programming instead of OOP which is the reason I expect to refer to other functions via closures instead of self-reference.

**Comment by jpaulomotta** [comment]
For now, this quick hack lets me call handlers inside withHandlers. Inspired by @istarkov solution and other answers:
```
compose(
    withHandlers({
        foo: props => () => {
           //Make magic
        }
    }),
    withHandlers({
        bar: ({foo}) => () => {
            foo()
        }
    })
)(Component)
```
props.foo() is accessible on the second withHandlers.

For simple use cases, it works fantastically, no need to create a clojure or do scope wizardry.

WreewanMorhee:
hope this feature will come true

rek:
@WreewanMorhee better to use Hooks now instead of this :)

PR #410, opened 2017-06-18
remove discord channel

ha404:
Clicked it and it led nowhere :(


wuct:
Thanks! The channel was defunct thus has been removed.

PR #412, opened 2017-06-23
Remove special cases from compose

jjant:
Removes extra logic to handle special cases from compose. Updated tests accordingly.

### DIFF
--- a/src/packages/recompose/compose.js
+++ b/src/packages/recompose/compose.js
@@ -1,11 +1,4 @@
 export default function compose(...funcs) {
-  if (funcs.length === 0) {
-    return arg => arg
-  }
-
-  if (funcs.length === 1) {
-    return funcs[0]
-  }
-
-  return funcs.reduce((a, b) => (...args) => a(b(...args)))
+  const identity = arg => arg
+  return funcs.reduce((a, b) => (...args) => a(b(...args)), identity)
 }
--- a/src/packages/recompose/compose.js
+++ b/src/packages/recompose/compose.js
@@ -1,4 +1,5 @@
+const identity = arg => arg
+
 export default function compose(...funcs) {
-  const identity = arg => arg
   return funcs.reduce((a, b) => (...args) => a(b(...args)), identity)
 }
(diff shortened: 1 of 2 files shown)


src/packages/recompose/compose.js
```diff
@@ -1,11 +1,4 @@
...
-  if (funcs.length === 0) {
-    return arg => arg
-  }
-
-  if (funcs.length === 1) {
-    return funcs[0]
-  }
-
-  return funcs.reduce((a, b) => (...args) => a(b(...args)))
+  const identity = arg => arg
```
istarkov: can be moved out of function body

TrySound:
Special cases are optiomisations. Let's keep it alive.

TrySound:
And which edge cases?

src/packages/recompose/compose.js
jjant: Right, thanks!

jjant:
@TrySound Are those small optimisations really worth it? I think that this version is a bit cleaner, there's no need to add if statements to cover cases that `reduce` already handles.

TrySound:
> Use to compose multiple higher-order components into a single higher-order component. This works exactly like the function of the same name in Redux, or lodash's flowRight().

TrySound:
With your changs it becomes different.

istarkov:
Just for one rare case

jjant:
I don't see it, when does it become different?

istarkov:
`expect(compose(fn)).toBe(fn)`

jjant:
In that case you get the identity function composed with `fn` (i.e, `arg => fn(arg)`), which is really the same thing, you just lose reference equality, but I don't see how that's a problem.

istarkov:
For me its also not a problem,  I think Ill accept this in near future.

istarkov:
Just played with simple perf
like
```
var fn0 = (a) => a + 1;
var b = performance.now(); var zzz= compose_old(fn0, fn0, fn0, fn0, fn0, fn0); for(let i =0; i< 1000000; ++i) zzz(1); console.log(performance.now() - b);
// vs
var b = performance.now(); var zzz= compose_new(fn0, fn0, fn0, fn0, fn0, fn0); for(let i =0; i< 1000000; ++i) zzz(1); console.log(performance.now() - b);
```
and it gives that even for case with 6 fn0, the perf difference is near 25%, old version is faster.
Can be an issue as I use (and as I know not only I) compose from recompose for a lot of other cases.

PR #415, opened 2017-06-28
Removed explicit Nothing.displayName in favor of the inferred name

Andarist:


istarkov:
break tests src/packages/recompose/__tests__/renderNothing-test.js

Andarist:
@istarkov sorry for this one, I had impression that ive ran the tests. Would u like to receive a PR with tests being configured (with [`husky`](npmjs.com/package/husky)) as prepush hook?

Ive fixed the failing test.

istarkov:
Yes support of lint-staged with husky will be cool, have you checked does this work if we compose withProps and renderNothing, will we have the same displayName as it was?

Andarist:
Gonna send husky PR in a minute.

As to testing this one more - even the changed test should be rather removed as its kinda testing JS feature (inferred names) and React internal way of [assigning a component's name](https://github.com/facebook/react/blob/b840229286ac2a82fa49553ce793cf7b953d1845/src/renderers/shared/stack/reconciler/ReactCompositeComponent.js#L224-L225).

Everything should work just fine with any composition as long as `displayName` is correctly set up for HOCs - but that has nothing to do with `Nothing` itself as its just a class component like any other.

PR #417, opened 2017-06-29
added with styles

dankram:
This allows for easy dynamic manipulation of component styles based on props.


dankram:
I still need to write tests which I would appreciate some help on.

istarkov:
Thank you.
Having withProps what the reason of this hoc? 
Also why styles property and not theme as like in themr or any other?

dankram:
As an example, withProps is just an extension on mapProps. I felt that people would also potentially use this a lot for dynamic styling. It may in fact be outside the scope of this project however.

TrySound:
I think this case is quite specific.

dankram:
That's fair. Thanks for taking a look!

PR #418, opened 2017-06-29
Added test task as prepush hook

Andarist:


wuct:
Why we need this feature?

Andarist:
Imho its a nice way to prevent people from pushing out broken code. There is still a way to push out even with broken tests if thats intended by someone, just by using `git push --no-verify`.

I've sent this PR after I've accidentally pushed out broken code and I've realized about it way later when I've checked the PR's CI status. Felt this could be prevented by failing fast on my machine - hence the PR.

wuct:
LGTM. @istarkov any thought?

istarkov:
As I understand `prepush` is used instead of `precommit` for situations like:
- Provide a commit with fail test example
- Provide a commit which will fix it.

So using `prepush` will allow to provide such PRs. IMO it's great we need to merge this.

istarkov:
Thank you!

Issue #328, opened 2017-03-07
memory leak of mapPropsStream at server side


istarkov:
Seems like current `mapPropsStream` could cause a memory leak at the server, and having that idea to call subscribe at `componentDidMount`  is not the thing what I want in most cases, 
the workaround I see is to use observable config like

```javascript
const config = {
  fromESObservable: Rx.Observable.from,
  toESObservable: stream => stream.take(1) // take(1) is a solution
}
```

As an example how take will help https://jsfiddle.net/2j2esxyL/7/

So having that there will be no call of componentWillUnmount at server side,
`take(1)` will close the stream so there will be no memory leaks.

Even the solution above will not solve all the cases, 
(the case at which mapPropsStream will wait something inside is not covered)
It seems like it will work for all cases I have internally.

I think if it's true (I haven't checked problem above) we need to add this into documentation.

@wuct ^

wuct:
Found a related issue https://github.com/facebook/react/issues/3714

wuct:
It seems there is not to much we can do without detecting the runtime environment.

> mapPropsStream will wait something inside

I am not sure what do you mean. Can you share an example?

wuct:
Maybe we can make two subscriptions, one in `componentWillMont` and  one in `componentDidMount`. The one in `componentWillMont` will unsubscribe immediately after `setState`.

```js
    componentWillMount() {
      // Subscribe to child prop changes so we know when to re-render
      const subscription = this.vdom$.subscribe({
        next: vdom => {
          this.setState({ vdom }, subscription. unsubscribe)
        }
      })
      this.propsEmitter.emit(this.props)
    }
```

The one in `componentDidMount` just works like how it should work and will be unsubscribed in `componentWillUnmount`.

istarkov:
Immediate unsubscription will cause cancellation of user created streams,  in all cases its not an expected behavior.
Will write an example of situation at which take(1) will not solve a problem later, now on mobile.

umidbekkarimov:
I guess `stream.take(1)` is only good solution for SSR now. We need `serverDidRender` hook (can't find link to discussion).

wuct:
Agree. I can't come up with other solutions.

PR #422, opened 2017-07-04
rxjs 5 tree shaking

rafaelkallis:
Currently, the rxjs config of recompose, for the `mapStreamToProps` and `componentFromStream` enhancers, imports the whole rxjs library. This is quite bad for size-sensitive builds. 

As described [here](https://github.com/ReactiveX/rxjs#installation-and-usage), it is possible to selectively import the features we need from rxjs in order to reduce the build size.

In this PR, I have changed the rxjs 5 config such that only necessary features get imported, dramatically decreasing build size for apps using the enhancers mentioned above with rxjs 5.

In my current app, I have reduced the production build size by **100KB** with this change.

### DIFF
--- a/src/packages/recompose/__tests__/setObservableConfig-test.js
+++ b/src/packages/recompose/__tests__/setObservableConfig-test.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import { mount } from 'enzyme'
+import 'rxjs/add/operator/map'
 import rxjs5Config from '../rxjsObservableConfig'
 import rxjs4Config from '../rxjs4ObservableConfig'
 import mostConfig from '../mostObservableConfig'
--- a/src/packages/recompose/rxjsObservableConfig.js
+++ b/src/packages/recompose/rxjsObservableConfig.js
@@ -1,7 +1,7 @@
-import Rx from 'rxjs'
+import { from } from 'rxjs/Observable/from'
 
 const config = {
-  fromESObservable: Rx.Observable.from,
+  fromESObservable: from,
   toESObservable: stream => stream,
 }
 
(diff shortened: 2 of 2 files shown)


rafaelkallis:
@acdlite question regarding the travis config, isn't a `yarn install` necessary? Or am I missing something obvious here? I believe that's the reason the travis checks are failing.

TrySound:
@rafaelkallis I guess it's because observable directory is not capitalized in this package.

istarkov:
Thank you.
Just from is useless, then users need to use bind operator or add other rx operators via global import. Everyone can create and use his own config if he need to reduce app size, or use smaller rx alternatives.

rafaelkallis:
@TrySound thanks for the tip, seems to work now.

@istarkov thanks for the feedback. The uselessness of only importing `from` is relative to the use-case. At least is should be mentioned in the docs that the `rxjsConfig` imports the whole rxjs lib and therefore is not suitable for size-sensitive builds. We could also make a separate `rxjsLiteConfig`. What are your thoughts?

istarkov:
I think that change of documentation is ok,
I thinks that we can place a link to gist with `rxjsLiteConfig` there too, as there are a lot of cases where we need an additional config, as an example SSR config to avoid memory leaks https://github.com/acdlite/recompose/issues/328#issue-212574406 
Having that every library can have similar problems we could get enormous amount of configs ;-) we need to support 
I'll approve such PR

wuct:
> Having that every library can have similar problems we could get enormous amount of configs ;-) we need to support

In that case, I think we should create a dedicated page for them.

rafaelkallis:
@istarkov sounds reasonable to me, PR can be closed in that case.

Issue #426, opened 2017-07-11
"trace" like component? 


alex-wilmer:
I use this all the time. it's just for dev, so maybe something that can be stripped in prod. but seems like a useful addition here. thoughts?

(non-recompose simple version)
```
export default C => p => (
  console.info(
    `%ctracing ${C.displayName} props:\n==========================`,
    'color: rgb(22, 138, 96);font-weight: bold;',
  ), // eslint-disable-line
  console.log(p), // eslint-disable-line
  <C {...p} />
);
```

alex-wilmer:
This is what I ended up with locally at the end of the day

```
/**
 *  Utility HOC designed to easily log received props.
*/

/* Recommended to be put on the global namespace in development mode:

global.trace = process.env.NODE_ENV === 'production'
  ? require(<path_to_trace>).noop
  : require(<path_to_trace>).default;

*/

/* usage:

  global.trace('Trying to fix issue #12')(Button)

  compose(
    global.trace() // string argument is optional.
  )

*/

export default (additionalMessage = '') => Component => props => (
  additionalMessage && console.log(additionalMessage),
  console.info(
    `%ctracing ${Component.displayName} props:\n${'tracing '.replace(
      /./g,
      '=',
    )}${Component.displayName.replace(/./g, '=')}=${'props:'.replace(
      /./g,
      '=',
    )}`,
    'color: rgb(22, 138, 96);font-weight: bold;',
  ),
  console.log(props),
  <Component {...props} />
);

export const noop = () => C => p => <C {...p} />;
```

timkindberg:
Seems pretty interesting to me.

alex-wilmer:
comment from colleague: "move noop branching choice inside the component via `branch`" which is a good idea. going to add it later

```
(additionalMessage = '') => branch(
  process.env.NODE_ENV === 'production',
  renderComponent(Component),
  Component => props => (/*...*/)
)
```


global.trace = require('@ncigdc/utils/trace');

wuct:
This feature is quire useful sometime, but I think consumers may want use their own ways to log. I suggest we provide a generalized HOC for all side-effects, like what Ramda's `tap()` or RxJS's `do()` does.

PR #428, opened 2017-07-12
Flow libdef + example recompose flow app

istarkov:
Here are
+ flow type lib definitions for recompose with tests.
+ example flow typed recompose app [Example App](http://grader-meets-16837.netlify.com/).
+ docs

The main idea that all you need is to provide type of enhancer in a form of:
```
const enhancer: HOC<*, EnhancedComponentProps> = compose(
```
All other types flow will infer automatically.

flow-typed PR: https://github.com/flowtype/flow-typed/pull/1032

See gif below:

PR #429, opened 2017-07-12
effect HOC (rxjs-like 'do') (#426)

alex-wilmer:
https://github.com/acdlite/recompose/issues/426

A passthrough HOC that doesn't do anything to the component. just a hook to inspect props or do some other effect.

Will updated the docs if you guys actually want this. I know I would use it a lot

### DIFF
--- a/src/packages/recompose/__tests__/effect-test.js
+++ b/src/packages/recompose/__tests__/effect-test.js
@@ -1,0 +1,19 @@
+import React from 'react'
+import { mount } from 'enzyme'
+import sinon from 'sinon'
+import { effect } from '../'
+
+test('effect calls its provided function', () => {
+  const component = sinon.spy(() => null)
+  component.displayName = 'component'
+
+  let external = ''
+
+  const Component = effect(({ foo }) => {
+    external = foo
+  })(component)
+
+  expect(Component.displayName).toBe('effect(component)')
+  mount(<Component foo="bar" />)
+  expect(external).toBe('bar')
+})
--- a/src/packages/recompose/effect.js
+++ b/src/packages/recompose/effect.js
@@ -1,0 +1,17 @@
+import setDisplayName from './setDisplayName'
+import wrapDisplayName from './wrapDisplayName'
+import createEagerFactory from './createEagerFactory'
+
+const effect = (fn = () => {}) => BaseComponent => {
+  const factory = createEagerFactory(BaseComponent)
+  const Effect = props => {
+    fn(props)
+    return factory(props)
+  }
+  if (process.env.NODE_ENV !== 'production') {
+    return setDisplayName(wrapDisplayName(BaseComponent, 'effect'))(Effect)
+  }
+  return Effect
+}
+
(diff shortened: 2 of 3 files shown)


istarkov:
The same as `withProps(props => undefined)`

jcheroske:
True, but more self-documenting.

alex-wilmer:
yeah technically this can be done with most of the existing HOC already

istarkov:
IMO not ;-)
Placing effects into render is not a good idea ;-)

alex-wilmer:
I'll leave my inspiration here https://github.com/NCI-GDC/portal-ui/blob/next/src/packages/@ncigdc/utils/trace.js

I originally wanted something that specifically logs to the console because it's such a common thing to do (inspect props at different parts of the composition stack).

jcheroske:
Perhaps a HOC that's only for debugging and doesn't run during production? We've all probably written one. Maybe it's time for the lib to support such a thing, idk...

istarkov:
The problem of trace above that it breaks `ref` behaviour as branch is functional component but internal `Component` can be not and vice versa

jcheroske:
Is it possible for the `trace` to look at the type of `BaseComponent` and return a matching wrapper? That might just be a slippery slope to hell.

istarkov:
BTW traces are useful only for short period of time, and leaving them in code is not a good idea, as some times later you will get enormous amount of traces.
I always use or withProps, or just add `console.log(anythingIWant) || code`
like 
```javascript
const myComp = (props) =>
  console.log(props) ||
  <div>Hello</div>;
```

istarkov:
BTW I think this PR #428 would allow me to reduce tracings - as now it will be possible to see prop types, as almost all of my traces is not about data itself but data structure

wuct:
To throw in my 2 cent, I sometime write `{console.log(props)}` inside
render functions :)

Ivan Starkov <notifications@github.com>於 2017年7月12日 週三，下午11:51寫道：

istarkov:
Also can be useful to use `debugger` statement
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/debugger

jcheroske:
Not to hijack this thread, but do you have any tips or resources for migrating to flow? I've been wanting to make the switch for some time now. I investigated typescript deeply a couple of months back, but abandoned it because it forced me to dump babel and eslint. I don't want to have to change tooling. Flow seems much more lightweight.

istarkov:
@jcheroske I've added a full working example of fully typed recompose code in that PR https://coolmenu-pguyikfzsr.now.sh/

especially decided to create something more hard than ordinary TODO list, to be sure that everything works fine
also I've added a tests for almost every usable HOC - and tests are also can be used as the source of how-to

timkindberg:
So it's sounding like a no to me. Just trying to wrap up the decision making process here. Sounds like there are several other fairly easy options for logging.

alex-wilmer:
Thanks for the feedback 😊

PR #431, opened 2017-07-17
Update API.md

PemL:
Documentation error: changing '_' -> { counter }

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -308,7 +308,7 @@ Example:
       decrementOn: ({ counter }) => (value) => ({
         counter: counter - value,
       }),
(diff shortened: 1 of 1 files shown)


wuct:
Actually this is not an error. `_` means we are not using this argument here.

PemL:
@wuct
Thank you for the clarification. I found it too be a bit ambigous, hence the reason for this pull request. 

I think some people, especially newer ones, will get confused. Or perhaps it's just me :).

wuct:
@PemL Hmmm, maybe what we need is some inline comment. How do you think?

PR #432, opened 2017-07-19
Fix withProps type def

tanem:
...since `withProps` accepts an object too.

Added a test but I'm a bit of a flow n00b, and didn't have much success running the tests with flow-typed to see if it was working as expected (think it's to do with the flow-type CLI hitting GitHub, will chase up there) 🤔


istarkov:
Cool, will look soon. Thank you.

istarkov:
Thank you!
Also here
https://github.com/flowtype/flow-typed/pull/1069

istarkov:
It's fun but I haven't knew that `withProps` can be used in non functional form ;-)

Issue #441, opened 2017-07-24
Type checking an enhancer that uses branch


tanem:
I'm trying to figure out how to sanely type an enhancer that uses `branch`. Example:

```js
// @flow

import React from 'react';
import ReactDOM from 'react-dom';
import {branch, compose, renderComponent, setDisplayName} from 'recompose';

import type {HOC} from 'recompose';

// --- button

type ButtonProps = {
  children?: React$Element<any>,
  onClick: () => mixed
};

const Button = ({children, onClick}: ButtonProps) =>
  <button onClick={onClick}>{children}</button>;

// --- anchor

type AnchorProps = {
  children?: React$Element<any>,
  url: string
};

const Anchor = ({children, url}: AnchorProps) =>
  <a href={url} target="_blank">{children}</a>;

// --- enhance

type EnhanceProps = {
  children?: React$Element<any>,
  // ?
  // url
  // onClick?
};

const enhance: HOC<*, EnhanceProps> = compose(
  setDisplayName('Button'),
  branch(
    ({url}) => Boolean(url),
    renderComponent(Anchor)
  )
);

const Enhanced = enhance(Button);

ReactDOM.render(
  // <Enhanced url="a url"}>anchor</Enhanced>,
  <Enhanced onClick={() => {}}>button</Enhanced>,
  document.getElementById('root')
);
```

So `Enhanced` renders as an `Anchor` if `url` is passed, otherwise it renders as a `Button`.

Setting up `EnhancedProps` with optional `url` and `onClick` won't work, one example where it doesn't is that you could then create a `Button` without an `onClick` and get no flow errors.

I'm getting the feeling that I should refactor this somehow, but I thought I'd ask here as I'm not sure if this is going to mean getting bitten again in future (we've used `branch` a fair bit, and are in the process of moving from `PropTypes` -> flow)?

istarkov:
https://github.com/acdlite/recompose/blob/master/types/flow-typed/recompose_v0.24.x-/flow_v0.49.x-/recompose_v0.24.x-.js#L213

istarkov:
btw in your case try to type EnhancedProps as like as
`type EP = type1 | type2`

istarkov:
where type1 and type2 are exact types like ` {|url: string|}`

tanem:
Thanks again @istarkov.

> btw in your case try to type EnhancedProps as like as
type EP = type1 | type2

> where type1 and type2 are exact types like {|url: string|}

[...]

PR #444, opened 2017-07-24
Expose Component type

tanem:
As per discussion in #441.


istarkov:
Thank you
also here
https://github.com/flowtype/flow-typed/pull/1069

PR #446, opened 2017-07-25
Update compose type def so it can handle more arguments

tanem:
Just following the existing pattern so the `compose` type def can handle a couple more HOCs. Not sure if there's a better way to be doing this 🤔


istarkov:
Please remove TODO section about compose at defs beginning

istarkov:
I think 9-10 is enough ;-)

tanem:
> Please remove TODO section about compose at defs beginning

Heh, I didn't even notice this at the top of the file 😆 All done.

wuct:
Looks good to me.

istarkov:
Guys please provide this change to flowtyped repo, as I cant do it now and for a few weeks

istarkov:
And @tanem thank you!

wuct:
Done https://github.com/flowtype/flow-typed/pull/1095

PR #451, opened 2017-07-26
Flatten only certain fields (optionally)

donavon:
For your consideration...

Adds an optional second parameter to `flattenProp`. If specified, only the fields specified in the array are flattened.

```js
flattenProp(
  propName: string,
  fields: ?Array<string>
): HigherOrderComponent
```

For example:
```js
const enhance = compose(
  withProps({
    object: { a: 'a', b: 'b' },
    c: 'c'
  }),
  flattenProp('object', ['a'])
)
const Ac = enhance(BaseComponent)

```
Base component receives props: { a: 'a', c: 'c', object: { a: 'a', b: 'b' } }

Added tests and updated docs.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -207,7 +207,8 @@ Renames multiple props, using a map of old prop names to new prop names.
 
 ```js
 flattenProp(
-  propName: string
+  propName: string,
+  fields: ?Array<string>
 ): HigherOrderComponent
 ```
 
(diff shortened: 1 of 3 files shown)


istarkov:
Thank you,
IMO `flattenProp` must be deprecated, as I see no benefit over the
`withProps(p => ({...p.myProp}))`

The problem of `flattenProp` as it can't be used well with `flow` type system we started to support
See https://github.com/acdlite/recompose/tree/master/types#part-2---other-hocs

istarkov:
Based on above I don't think that adding additional string argument is a good idea.

IMO this is better as more readable
```
withProps(({myObj: {a, b, c}}) => ({a, b, c}))
```

donavon:
Agreed!

PR #459, opened 2017-07-31
update recompose-relay to use relay-modern v2

graingert:


graingert:
This is https://github.com/acdlite/recompose/pull/340 rebased with support for react-relay 0.x or 1.x

graingert:
https://www.npmjs.com/package/react-relay-compat

https://unpkg.com/react-relay-compat@0
https://unpkg.com/react-relay-compat@1

graingert:
you can pull in the modern version with:

```js
const MaybeRelayModern = require('react-relay-compat/modern');
 
if (MaybeRelayModern) {
  const RelayModern = MaybeRelayModern;
  ...
}
```

istarkov:
Sorry for the delays with recompose-relay PRs. I'll try to find a time next week to play with relay modern so will not be really dumb looking into such PRs. As now I have zero opinion what to do with this, sorry.

graingert:
@istarkov this is the minimum change needed to support relay modern (or relay) without the app just crashing when you launch it. Can you release a -alpha.1 with this in?

istarkov:
Ok, when I'll get a publishing permissions for recompose-relay I'll publish prev version at 0.4.0 then merge this and publish as 0.5.0

graingert:
@istarkov ah, @acdlite hasn't given you publish permissions?

istarkov:
I've just asked for it ;-)

graingert:
@istarkov I've added you to the recompose group on npm. We can create @recompose/core and @recompose/relay packages.

istarkov:
After using modern relay at work I see no reason to support recompose for relay modern. I'll add a deprecation notice into a package.

It's easy to reorder relay arguments as

```javascript
import { createFragmentContainer } from 'react-relay';
export default query => base => createFragmentContainer(base, query);
```
 or use use one liner with ramda

```
curry(createFragmentContainer)(__, query),
```

the same for other containers.

Having that all recompose-relay is just one function I recommend to copy it into your source.

PR #462, opened 2017-08-02
Fix a typo in types/README.md

felixonmars:


istarkov:
Thank you, its strange that with my english here is just one typo :-)

PR #463, opened 2017-08-03
Simpler classful component check (fixes #458)

### LINKED ISSUE #458: Compatibility with non-React VDOMs (e.g. Preact, Inferno)
- [x] I have searched for similar issues, but found none.
 - [x] I have read the `README`.
 - [x] I attempted to read the `CONTRIBUTORS` guide, but found none.

I am wondering if there is interest within this project to support other VDOM implementations such as `preact` and `inferno`. I am currently using `recompose` with `preact` and I believe it adds a lot of value to our project; but there are bugs caused by accessing non-public React API's such as the use of [`Component.prototype.isReactComponent` in `isClassComponent`](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/isClassComponent.js#L5).

If this is considered valuable for `recompose` I would be happy to contribute towards that direction. Thanks for the quality library.

developit:
Instead of checking for the `isReactComponent` property (which is internal to React), it seems advantageous to check for `render()`. This improves Preact support (when using recompose with non-preact-compat components) but also importantly works with components that don't inherit from `Component`.

### DIFF
--- a/src/packages/recompose/isClassComponent.js
+++ b/src/packages/recompose/isClassComponent.js
@@ -2,7 +2,7 @@ const isClassComponent = Component =>
   Boolean(
     Component &&
       Component.prototype &&
-      typeof Component.prototype.isReactComponent === 'object'
+      typeof Component.prototype.render === 'function'
   )
 
 export default isClassComponent


billneff79:
To maintain compatibility with what he is already doing in `isClassComponent`, it would have to return `true` for `React.Component` but `false` for `React.PureComponent` and `React.AsyncComponent`, since `React.Component` is the only one that has the `isReactComponent` prop.  All of them would have `render`, but only `React.Component` has `setState` and `forceUpdate`.  If he doesn't care about changing it so it returns `true` for `PureComponent` then checking for `render` would work as well.

billneff79:
^ Just saw your comment about being compatible with things that are classes but don't extend `React.Component`.  Again goes back to what kind of contract he is going for with `isClassCompoenent` - is it just trying to return something that is not a functional component (in which case checking for `render` is perfect), or is it trying to check only that it is extending `React.Component`

istarkov:
Thank you

gaearon:
In case other library authors see this PR and adopt the same approach, I’m encouraging you not to.

As noted in https://github.com/acdlite/recompose/pull/538#issuecomment-334820814, this change was actually a major change (and broke people’s code in production). It can break code because `render` might be on the instance rather than on the prototype in some cases. It can also break code in React Native.

Here is the exact algorithm React is using:

* Check for `fn.prototype && fn.prototype.isReactComponent`
* If it is truthy
  * This is a class component and must be `new`'d
* If it is falsy
  * Call the function without `new`
  * If the return value is an object and has a `render` method on it
    * Treat it as a class instance
  * Otherwise, treat it as a functional component

React plans to continue using this heuristic. In this sense, if for some reason your library needs to know whether a component is "newable" or not, `fn.prototype && fn.prototype.isReactComponent` is a safe way to check it.

I understand we might not be able to persuade Preact or others to adopt the same heuristic, but I hope this is helpful.

PR #471, opened 2017-08-10
Update API.md

mechanical-turk:
not sure if this was an error, but I believe compose returns a function that is a HOC itself, so it would need to take in a component, right?


istarkov:
It returns hoc, this is written in docs

PR #473, opened 2017-08-14
Use eager factory optimization only in production env (close #442)

### LINKED ISSUE #442: should `isReferentiallyTransparentFunctionComponent` be completely off for non-production env?
Hi.

Right now

```js
(process.env.NODE_ENV === 'production' || !Component.propTypes)
```

from [here](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/isReferentiallyTransparentFunctionComponent.js) leads to lack of actual element while testing with Enzyme (`NODE_ENV=test`) so very dummy stateless components can't be found in a tree. Same problem with react-dev-tools (`NODE_ENV=development`), I can't see the component because function was directly invoked with props.

I understand the purpose of `createEagerFactory`, and the idea is really nice, but in my opinion it should be production only optimization. Should we consider removing `|| !Component.propTypes` part from the condition or something like this? It's really confusing and took a lot of time to find the reason.

deepsweet:
As discussed in #442.

* We don't care about `propTypes` check anymore since it's already no-op in prod env.
* Notice few `shallow` → `mount` changes in tests – that's actually the confusion which I described in corresponding issue.

:v:

### DIFF
--- a/src/packages/recompose/isReferentiallyTransparentFunctionComponent.js
+++ b/src/packages/recompose/isReferentiallyTransparentFunctionComponent.js
@@ -2,9 +2,9 @@ import isClassComponent from './isClassComponent'
 
 const isReferentiallyTransparentFunctionComponent = Component =>
   Boolean(
-    process.env.NODE_ENV === 'production' &&
-      typeof Component === 'function' &&
+    typeof Component === 'function' &&
       !isClassComponent(Component) &&
+      !Component.propTypes &&
       !Component.defaultProps &&
       !Component.contextTypes
   )
--- a/src/packages/recompose/isReferentiallyTransparentFunctionComponent.js
+++ b/src/packages/recompose/isReferentiallyTransparentFunctionComponent.js
@@ -4,7 +4,6 @@ const isReferentiallyTransparentFunctionComponent = Component =>
   Boolean(
     typeof Component === 'function' &&
       !isClassComponent(Component) &&
-      !Component.propTypes &&
       !Component.defaultProps &&
       !Component.contextTypes
   )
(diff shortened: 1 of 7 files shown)


wuct:
The title of this PR and the commit should be "Use eager factory optimization only in production env ".

deepsweet:
@wuct oops, fixed.

wuct:
Looks good to me, thanks!

wuct:
Hmmm, actually I think we should move the `env` check out of `isReferentiallyTransparentFunctionComponent()`, otherwise there is an abstraction leak.

deepsweet:
Agreed. So both to `createEagerFactory` and `createEagerElement`?

deepsweet:
Hm, looks like `createEagerElementUtil ` is the right place.

deepsweet:
Done.

I returned check for `propTypes` back because it's actually a part of abstraction, so this way it's not related to any env and concepts are divided. But optimization will not work even in production env if you have `propTypes`.

Options:
1. Remove check for `propTypes` from `isReferentiallyTransparentFunctionComponent` at all.
2. I'd like to work on something similar to [babel-plugin-transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) but for `setPropTypes`, I think it's a more proper way to achieve prod-only optimization.

WDYT?

src/packages/recompose/isReferentiallyTransparentFunctionComponent.js
```diff
@@ -4,9 +4,9 @@ const isReferentiallyTransparentFunctionComponent = Component =>
   Boolean(
     typeof Component === 'function' &&
       !isClassComponent(Component) &&
+      !Component.propTypes &&
```
istarkov: seems like break previous behaviour?

istarkov:
IMO option 1 is preferred, having that it is fully compatible with previous behaviour.

deepsweet:
Done.

deepsweet:
`isReferentiallyTransparentFunctionComponent returns true for functions that use propTypes`

istarkov:
Thank you

PR #474, opened 2017-08-14
Add componentFromStream and createEventHandler flow definitions

TrySound:


types/flow-typed/recompose_v0.24.x/flow_v0.49.x-/recompose_v0.24.x.js
```diff
@@ -332,4 +332,10 @@ declare module 'recompose' {
   ): Component<A>
 
   declare export function hoistStatics<A, B, H: HOC<A, B>>(hoc: H): H
+
+  declare export function componentFromStream<T>(
+    (props$: *) => $Exact<*>
```
istarkov: why existential why exact?
also must be aligned with mapPropsStream where any is used for props

TrySound: Without exact test fails.

istarkov:
Thank you

PR #476, opened 2017-08-15
Allow to use withStateHandlers with SyntheticEvents

### LINKED ISSUE #456: How do you get at events in withStateHandlers?
Here's a contrived example (I'm using React Native).
```
const App = ({ positionX, handleButtonPressX, positionY, handleButtonPressY }) =>
  (<View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
    <Button title="Set X Position" onPress={handleButtonPressX} />
    <Button title="Set Y Position" onPress={handleButtonPressY} />
    <Text style={{ marginTop: 30 }}>
      {`X: ${positionX}\nY: ${positionY}`}
    </Text>
  </View>);

const trackPositionX = compose(
  withState('positionX', 'setPositionX', ''),
  withHandlers({
    handleButtonPressX: ({ setPositionX }) => (event) => {
      setPositionX(Math.floor(event.nativeEvent.pageX));
    },
  }),
);

const trackPositionY = withStateHandlers(
  () => ({
    positionY: '',
  }),
  {
    handleButtonPressY: () => (event) => ({
      positionY: Math.floor(event.nativeEvent.pageY),
    }),
  },
);
```

When clicking the *Set Y Position* button I get the error  
 `null is not an object (evaluating 'event.nativeEvent.pageY')`

It seems  `event` is undefined. I know this is probably a basic programming concept I'm not getting. Can someone help me understand?

Thanks!

istarkov:
Fixes #456 

Problem example
https://codesandbox.io/s/DRQ8y0zBx

Console output 
```
Warning: This synthetic event is reused for performance reasons. If you're seeing this, you're accessing the property `target` on a released/nullified synthetic event. This is set to null. If you must keep the original synthetic event around, use event.persist(). See https://fb.me/react-event-pooling for more information.
```

### DIFF
--- a/src/packages/recompose/__tests__/withStateHandlers-test.js
+++ b/src/packages/recompose/__tests__/withStateHandlers-test.js
@@ -4,6 +4,40 @@ import sinon from 'sinon'
 
 import { compose, withStateHandlers } from '../'
 
+test.only('withStateHandlers should persist events passed as argument', () => {
+  const component = ({ value, onChange }) =>
+    <div>
+      <input type="text" value={value} onChange={onChange} />
+      <p>
+        {value}
+      </p>
+    </div>
+
+  const InputComponent = withStateHandlers(
+    { value: '' },
+    {
+      onChange: () => e => ({
+        value: e.target.value,
+      }),
+    }
+  )(component)
+
+  const wrapper = mount(<InputComponent />)
+  const input = wrapper.find('input')
+  const output = wrapper.find('p')
+  // having that enzyme simulate does not simulate real situation
+  // emulate persist
+  input.simulate('change', {
+    persist() {
+      this.target = { value: 'Yay' }
+    },
+  })
+  expect(output.text()).toBe('Yay')
+
+  input.simulate('change', { target: { value: 'empty' } })
+  expect(output.text()).toBe('empty')
+})
+
 test('withStateHandlers adds a stateful value and a function for updating it', () => {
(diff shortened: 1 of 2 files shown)


istarkov:
And yes it looks like a very dirty hack but the only other option I see is to use non functional form of setState. 
With non functional form we will lost atomic updates so not a good option.

(Also it's possible to copy event properties `newEvt = {...oldEvt}` BTW having that persist exists It's better to use it)

jbovenschen:
Encountered the same issue, also needed to call the event.persist() function before passing it back to the top. Good fix which looks a little bit hacky though.

mnpenner:
@wuct 0.25.0 is out but this doesn't appear to be in it, and 0.24.1 doesn't exist. Do you have an ETA on when this will land on npm?

wuct:
@mnpenner I believe it landed in [0.25.0](https://github.com/acdlite/recompose/releases/tag/v0.25.0). Have you tried it?

mnpenner:
@wuct Pretty sure I did try it... will try again when I find some time, but I swear I checked and the event properties weren't available.

wuct:
@mnpenner You can see it works in this [jsfiddle](https://jsfiddle.net/wuct/q74todh6/2/).

mnpenner:
@wuct I might have messed up my implementation somehow. Everything seems to be fine now :man_shrugging: Thanks for your help

PR #487, opened 2017-08-21
Updated jsfiddle example base link. Fixes #486

### LINKED ISSUE #486: Replace older jsfiddle link with updated example
The jsfiddle linked from the README.md is based on the older React jsfiddle base. Since that base was created jsfiddle properly supports babel+jsx in the editor directly, which includes syntax highlighting.

The pull request to update to a new fiddle would be trivial, but I'm not certain of some possible opinions. The current fiddle uses `compose`, `withState`, and `withHandlers`, which has been surplanted by `withStateHandlers`. I created a [demo using `withStateHandlers`](https://jsfiddle.net/samsch/p3vsmrvo/). To support `withStateHandlers`, I used latest through unpkg ([https://unpkg.com/recompose/build/Recompose.js](https://unpkg.com/recompose/build/Recompose.js)) rather than pointing at a specific version.

However, if a different demo, or the older demo is preferred, its easy to (re)build off of my [updated base React jsfiddle](https://jsfiddle.net/samsch/nxjd9ens/), and gain syntax highlighting support.

samsch:
New jsfiddle uses built-in babel mode with syntax highlighting, as mentioned in #486.

This new fiddle uses the latest version of Recompose, and the example uses `withStateHandlers` rather than `compose`, `withState`, and `withHandlers`.

Code formatted externally with Prettier (default config, using online sandbox).

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Recompose is a React utility belt for function components and higher-order compo
 
 [**Full API documentation**](docs/API.md) - Learn about each helper
 
-[**Recompose Base Fiddle**](https://jsfiddle.net/acdlite/69z2wepo/41596/) - Easy way to dive in
+[**Recompose Base Fiddle**](https://jsfiddle.net/samsch/p3vsmrvo/2/) - Easy way to dive in
 
 ```
 npm install recompose --save


istarkov:
Thank you!!!

PR #494, opened 2017-08-25
Add hoverable hoc

caesay:
This is a higher order component which will wrap a base component with an event handling component and provide it's current mouse hover state to the base component.


istarkov:
Hi, Thank you. 
We prefer to not add into recompose highly specialized enhancers like yours.
One of the reasons that the idea of recompose is to provide you with some base using it you can create your own Components or enhancers. 
The other reason that everyone have a lot of ideas for enhancers he need, and we just can't add and support all of them.
I also use some highly specialized enhancers internally - like size detection, data manipulation, etc.

You can create your own enhancers library or use already existing like
https://github.com/deepsweet/hocs and others.

PR #517, opened 2017-09-15
hoist-non-react-statics@2.3.1

ahutchings:
Updates hoist-non-react-statics to version 2. This brings the version range in line with react-redux's dependency on hoist-non-react-statics, allowing build tooling to dedupe and produce smaller bundles.


wuct:
Thanks!

PR #518, opened 2017-09-16
Clarify withStateHandlers behavior

### LINKED ISSUE #515: withStateHandlers documentation does not agree with behavior
The `withStateHandler` docs say:
> Every state updater function ... must return a new state or undefined

[But the code uses React's `setState`](https://github.com/acdlite/recompose/blob/1465683bcfc58b2939d246e7a30b760ac192c977/src/packages/recompose/withStateHandlers.js#L25)

[React's `setState` docs](https://facebook.github.io/react/docs/react-component.html#setstate) say:
> The output of the updater is shallowly merged with `prevState`.

idris:
Closes #515

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -291,7 +291,7 @@ withStateHandlers(
 Passes state object properties and immutable updater functions
 in a form of `(...payload: any[]) => Object` to the base component.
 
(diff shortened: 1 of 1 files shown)


istarkov:
Thank you!

Issue #520, opened 2017-09-19
Export createEventHandlerWithConfig()?


acchou:
The documentation suggests using functions like `componentFromStreamWithConfig()` instead of `componentFromStream()` to avoid having to set global state with `setObservableConfig()`.

It seems there's also a similar situation with `createEventHandlerWithConfig()` vs `createEventHandler()`. There are two issues though:

(1) There's no mention of `createEventHandlerWithConfig()` in the docs.
(2) The Typescript typings for recompose don't export `createEventHandlerWithConfig()`, so it appears to be unavailable as an import.

I'd be happy to submit a pull request if I get some direction on where this should go.

wuct:
Thanks for reporting. Tasks we can do including:

1. mention `createEventHandlerWithConfig()` in the doc.
2. re-export `createEventHandlerWithConfig()` in [index.js](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/index.js)

A PR is welcome :)

We don't have TypeScript def in this repo, so you probably need to open a PR where it is.

PR #522, opened 2017-09-19
export `createEventHandlerWithConfig()` and add documentation.

acchou:
Address issue #520. I'll also suggest a separate PR for typescript types in the appropriate repo.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -889,12 +889,27 @@ const Timer = enhance(({ timeElapsed }) =>
 
 ```js
 createEventHandler<T>(): {
-  handler: (value: T) => void
-  stream: Observable<T>,
+  handler: (value: T) => void,
+  stream: Observable<T>
 }
 ```
 
 Returns an object with properties `handler` and `stream`. `stream` is an observable sequence, and `handler` is a function that pushes new values onto the sequence. Useful for creating event handlers like `onClick`.
+
+### `createEventHandlerWithConfig()`
+```js
+createEventHandlerWithConfig<T>(
+  config: {
+    fromESObservable<T>: ?(observable: Observable<T>) => Stream<T>,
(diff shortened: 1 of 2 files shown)


docs/API.md
```diff
@@ -889,13 +889,28 @@ const Timer = enhance(({ timeElapsed }) =>
...
-  handler: (value: T) => void
-  stream: Observable<T>,
+  handler: (value: T) => void,
+  stream: Observable<T>
 }
 ```
 
 Returns an object with properties `handler` and `stream`. `stream` is an observable sequence, and `handler` is a function that pushes new values onto the sequence. Useful for creating event handlers like `onClick`.
 
+### `createEventHandlerWithConfig()`
```
wuct: Could you also add an entry to the TOC section?

wuct: It turns out that `mapPropsStreamWithConfig` is also missing from the TOC section.

docs/API.md
```diff
@@ -889,13 +889,28 @@ const Timer = enhance(({ timeElapsed }) =>
...
+```js
+createEventHandlerWithConfig<T>(
+  config: {
+    fromESObservable<T>: ?(observable: Observable<T>) => Stream<T>,
+    toESObservable<T>: ?(stream: Stream<T>) => Observable<T>,
+  }
+) => (): {
+  handler: (value: T) => void
```
wuct: missing a `,` here.

docs/API.md
acchou: Added these functions to the TOC, and also `componentFromStreamWithConfig()`.

acchou:
I noticed that I had to npm install `change-emitter` and `hoist-non-react-statics` in order to push. Should these deps be in `package.json`?

wuct:
@acchou Thanks! And yes, we can put them in `devDependencies`.

PR #523, opened 2017-09-20
Use $Compose instead Compose

istarkov:
Fixes almost all flow typing issues we had!!!!!
Waiting for flow-bin to be updated.


types/flow-typed/recompose_v0.24.x/flow_v0.55.x-/recompose_v0.24.x.js
```diff
@@ -64,64 +64,6 @@ declare module 'recompose' {
 
   declare type UnaryFn<A, R> = (a: A) => R
 
-  declare type Compose = (<A, B, C, D, E, F, G, H, I>(
```
tanem: Ha, nice to see this go, we hit the limit again not long after I submitted the earlier PR to extend it 😅

types/flow-typed/recompose_v0.24.x/flow_v0.55.x-/test_mapProps.js
```diff
@@ -26,7 +26,7 @@ const enhacer: HOC<*, EnhancedCompProps> = compose(
   // seems like this https://github.com/facebook/flow/issues/4342 issue
   withProps(props => ({
     a: (props.a: string),
-    // Must $ ExpectError but not
+    // $ExpectError but not
```
wuct: If it does throw error, we can remove `but not` as well.

PR #529, opened 2017-09-22
Typo

chinmaymahajan:
- [x] Fixed the typo


istarkov:
Its not a typo, it does not map over props. It maps enhanced (owner) props into base component props.

PR #531, opened 2017-09-26
Add link to a HOCs criticism

istarkov:
### I haven't got this part: 

`Additionally, the composition model here is dynamic! Everything happens inside of render, so we get to take full advantage of the React lifecycle and the natural flow of props & state.`

### I don't agree with 

`flow` part about `you’ll probably find it much easier to write a type definition for your component with a render prop`

### Partially agree with:

- Indirection. We don’t have to wonder where our state or props are coming from. We can see them in the render prop’s argument list.

(_We can also see in HOCs, as most(all?) HOCs here contain some kind of mapper functions_)
(_Render props can cause similar problems as callback hell_)

### Agree with but haven't seen this often in real practice:

- Naming collisions. There is no automatic merging of property names, so there is no chance for a naming collision

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,12 @@ Recompose focuses on another unit of composition: **higher-order components** (H
 
 Recompose provides a toolkit of helper functions for creating higher-order components.
 
-## [Should I use this? Performance and other concerns](docs/performance.md)
+## Should I use this?
+
+[Performance and other concerns](docs/performance.md)
+
+
+[Use a Render Prop](https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce) [Michael Jackson](https://twitter.com/mjackson) article about whats wrong with HOCs.
 
 ## Usage
 


README.md
```diff
@@ -195,7 +195,12 @@ Recompose focuses on another unit of composition: **higher-order components** (H
 
 Recompose provides a toolkit of helper functions for creating higher-order components.
 
-## [Should I use this? Performance and other concerns](docs/performance.md)
+## Should I use this?
+
+[Performance and other concerns](docs/performance.md)
+
+
+[Use a Render Prop](https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce) [Michael Jackson](https://twitter.com/mjackson) article about whats wrong with HOCs.
```
NMinhNguyen: nit: missing an apostrophe in `what's` :sweat_smile:

istarkov: In my native language apostrophe is needed for only one word - D'Artagnan, all other Russian words dont use it :-)
Thank you!!!

NMinhNguyen: Я прекрасно понимаю и не имел плохих намерений - просто решил подсказать, как правильно :smile: 

In English: I didn't mean to offend, just wanted to suggest the correct spelling haha

istarkov: Hahahaha ;-)
I know, it's just a fun joke 

- why in Russian alphabet we have apostrophe.
- without apostrophe we can't write one of the greatest Russian words - D'Artagnan

(here he is a hero of anecdotes and very popular)

wuct:
> Additionally, the composition model here is dynamic! ...

Probably something we can do with `withState`, `mapProps` and `branch`.

mcky:
The dynamic composition part I'm also struggling to grasp the benefits of. They've definitely gone crazy for it though, with react-router's controversial rewrite..

Issue #533, opened 2017-09-27
Fiddle is not working with React 16


fatso83:
The distributable has moved. See the console.output
https://jsfiddle.net/samsch/p3vsmrvo/2/

```
17:18:08.043 unpkg.com/react@16.0.0/dist/react.js Failed to load resource: the server responded with a status of 404 ()
17:18:08.043 unpkg.com/react-dom@16.0.0/dist/react-dom.js Failed to load resource: the server responded with a status of 404 ()
17:18:08.044 unpkg.com/recompose@0.25.1/build/Recompose.js:7 Uncaught TypeError: Cannot use 'in' operator to search for 'default' in undefined
    at unpkg.com/recompose@0.25.1/build/Recompose.js:7
    at React__default (unpkg.com/recompose@0.25.1/build/Recompose.js:4)
    at unpkg.com/recompose@0.25.1/build/Recompose.js:5
17:18:08.045 babel.js:60934 You are using the in-browser Babel transformer. Be sure to precompile your scripts for production - https://babeljs.io/docs/setup/
runScripts @ babel.js:60934
17:18:08.102 Inline Babel script:18 Uncaught TypeError: withStateHandlers is not a function
    at <anonymous>:26:13
    at run (babel.js:60802)
    at check (babel.js:60868)
    at loadScripts (babel.js:60909)
    at runScripts (babel.js:60936)
    at transformScriptTags (babel.js:324)
````

fatso83:
Working https://jsfiddle.net/oligofren/p3vsmrvo/22/

istarkov:
PR?

istarkov:
fixed, thank you

samsch:
@fatso83 Whoops, I didn't see this issue and your fiddle before creating my pull. I noticed my other fiddles breaking and remembered the one here would be broken too.

Sorry for (unintentionally) ignoring you.

istarkov:
The most fun thing that you both found it in almost same time ;-)
May be you are twins?

PR #534, opened 2017-09-27
Updated base fiddle link

samsch:
Base fiddle was using a unpkg link pointed at the old (React 15) dist url but without a React version, so broke with the release of React 16. Now points to the correct location for React 16, with 16 specified (shouldn't break this way again).


istarkov:
closes #533

istarkov:
Thank you

PR #538, opened 2017-10-02
Remove createEagerFactory and createEagerElement

### LINKED ISSUE #489: [0.25.0 update] createEagerFactory production only error
after update to 0.25.0 start getting 
```
Error walking your react tree
TypeError: Class constructor OnMount cannot be invoked without 'new'
at createEagerElementUtil (/app/node_modules/recompose/utils/createEagerElementUtil.js:18:12)
at /app/node_modules/recompose/createEagerFactory.js:18:49
at ShouldUpdate.render (/app/node_modules/recompose/shouldUpdate.js:45:16)
at /app/node_modules/react-tree-walker/commonjs/index.js:148:35
at doTraverse (/app/node_modules/react-tree-walker/commonjs/index.js:78:21)
```
was all fine on 0.24.0 

OnMount component is pretty plain 
```
export default (timeout = 0) => BaseComponent =>
  class OnMount extends PureComponent {
    state = { isMounted: false };

    componentDidMount() {
      this.mounted = true;
      setTimeout(() => {
        if (this.mounted) {
          this.setState({ isMounted: true });
        }
      }, timeout);
    }

    componentWillUnmount() {
      this.mounted = false;
    }

    mouted = false;

    render = () =>
      <BaseComponent isMounted={this.state.isMounted} {...this.props} />;
  };

```

deepsweet:
closes #489, as discussed.


istarkov:
Super, wiil publish today-tomorrow!

istarkov:
First time in my life I see usage of react createFactory in a wild

deepsweet:
well I just tried to have the minimal diff :) we can go with JSX instead.

istarkov:
The minimal diff is fine for me, with JSX I think it will be a problems with preact etc

wuct:
But there is no `createFactory` in preact, I think we should use JSX instead.

istarkov:
@wuct https://github.com/developit/preact-compat

wuct:
@istarkov Thanks! This PR looks good to me now :)

istarkov:
Thank you!!!!!

oliviertassinari:
I'm curious about the motivation behind this change. Are you saying it's playing with fire forcing this performance optimisation, when instead, it could be something done by React internally?
Also, I never understood why this performance optimisation has been enable for production only at some point. Why not keeping it for all the env? Thanks for your 💡.

istarkov:
Initially all began here #442 (PR #473), at that time optimisation already worked differently at production and development (but for PropTypes case) and also was a point of problems with enzyme tests. But just that change wasn't the point of following problems ;-) 
In that days we also merged this [PR](https://github.com/acdlite/recompose/pull/463/files) where instead of undocumented check is something is a React Component we started to check on `render` function existence on prototype. Since then our problems become more powerful - we found that not all the React Native components has render method in prototype, that some guys binds render method (so it does not exist in prototype) and all of them got exception only at production.
So the possible solutions was 
- make optimisation to work in dev also - that does't solve the error - people will get strange error in dev also ;-)
- return previous check - breaking support of React-like frameworks and using undocumented (not sure it exists in react 16) method
- found a better way to detect that something is Class Component and not the function (we didn't found it) (this solves almost all above problems but not all)
- remove optimisation - as having that we can do it - why React itself can't do it also (I think React don;t do it before because it had functional components internally based on class components - but not sure I'm right here)

Having that react vdom is very fast - and having that at dev I see no performance problems last month+ (I know that it is very bad test - "I see no perf problem") we decided to remove it.
May be it will cause new discussion and proposals with new better solution ;-)

istarkov:
I've asked @acdlite here https://twitter.com/icelabaratory/status/916360249635999744
and will repeat here 
@acdlite can you explain why React doesn't use eager optimisations like you did in recompose?

istarkov:
image

oliviertassinari:
@istarkov Thanks a lot for taking the time to answer the question in detail! Keep the good work.

NMinhNguyen:
Shouldn’t this public API removal be considered a breaking change? Someone could be relying on these factories in their own HOC implementations.

deepsweet:
minor version bump while major is zero is already a breaking change according to semver.

NMinhNguyen:
My bad, should’ve been clearer - perhaps having the word BREAKING in the release notes would’ve been clearer, but maybe it’s implicit from the minor version increase 😅

kof:
@istarkov  Was a different option for eager optimizations discussed? For e.g. only supporting functional components?

gaearon:
>return previous check - breaking support of React-like frameworks and using undocumented (not sure it exists in react 16) method

Looking for `isReactComponent` on the prototype **is the supported way** to check for whether something is a React component class. No, there are no plans to change it, and you can adopt the same check in your library.

Checking for `render` on the prototype is not, and the reasons you discovered later (with people’s code breaking in production) are exactly why we weren’t doing it in React either.

@developit @istarkov Please feel free to reach out to us next time you need to check something like this. We’re happy to explain motivations for why went for a particular check.

istarkov:
@kof the main issue was our inability to distinguish functional component from class component, both are functions (class is still a constructor function), both can have no render function in prototype, there were non documented method for React (now documented `isReactComponent`) but not for Preact. 
Also even this optimization  reduces virtual tree size there are no any performance tests showing that this is somehow seriously improves performance in real life. I haven't seen any drawbacks in our production or even development code.

istarkov:
And also this ;-)
image

kof:
@istarkov I guess it depends on how it is used. I am right now investigating what to do with recompose, because it was used very extensively and I see about 10HOCs around a component. This is neither nice to look at in dev tools nor it can be inexpensive at scale in a large application.

I know one could "just be careful" about how many recompose functions which create a HOC to use, but this is hard to enforce at scale and I would rather look for a different solution than to relying on a guideline.

kof:
As there is a documented way to detect a class (thanks @gaearon for pointing out), we should find a way to use it. If preact doesn't support it, it is a problem in preact and can be easily fixed. We should not punish every one else.

gaearon:
In general I agree that if Recompose can optimize it, it would be nice if it did. There is extra fixed cost associated with each layer. It’s not much, but it accumulates especially in projects that heavily rely on HOCs.

istarkov:
@gaearon but in general such optimization can be done on React size - if component is functional, have no `defaultProps`, and `propTypes` you can call it as a function on React size without creating an element?

The other problem that this optimisation is different on production and dev, as you have no propTypes on production but have on dev (so you must createElement in such case)

istarkov:
@kof If I had no pack of issues about this - different tree structure at dev and prod, preact support etc I could say you are right. But I had. 
Since we removed this feature such issues are gone. 
This feature didn't work in all cases - just in a series of simple hocs like `withProps`, `mapProps` etc, any class based HOC broke that behavior. Also there are no tests that it's much more faster, and a lot of questions how it will work with any new React feature. Does `React.forwardRef` returns function or class etc. 
There are plain version of recompose available https://github.com/neoziro/recompact which removes all the tree depth and was fully compatible with recompose.
There were even tests written for compare recompose, recompact etc - yes in synthetic tests - 10x withProps in one compose, recompose is worse but not much.

kof:
I haven't looked yet into each issue, but
> different tree structure at dev and prod 

yeah should not happen, this optimization should be always in place if at all

> preact support

If it is about adding isReactComponent I am sure @developit will accept a pr for this.

deepsweet:
what would be (and it actually was) really misleading to many people is that you can't find something in React elements tree using Enzyme or similar tools because it was automagically "squashed" by quite internal and advanced condition check.

dev tools prettiness is doomed in any way because we still have stateful HOCs even if we can just call the rest as functions.

kof:
What about a separate function which does the optimizations?

deepsweet:
it may play well at least for me if it's easy to alias it using some Babel plugin or Webpack only in production mode, just like react → preact seamless transition.

gaearon:
Here's something I don't understand.

I agree that in general the edge case of "executing" arbitrary *user-provided* components can be too confusing.

But can't Recompose safely collapse *is own* components? E.g. `mapProps(mapProps(mapProps(...)))` should be fairly easy and safe to "collapse" in production—is it not?

deepsweet:
>should be fairly easy and safe to "collapse" in production

and we made it so https://github.com/acdlite/recompose/pull/473, but then it led to some other issues like https://github.com/acdlite/recompose/issues/489 but I don't actually recall all the details 😳

kof:
> My main frustration is that it runs ok in development while crushes on production.

yeah, its dangerous to have different behavior in production, much better to stick with it everywhere.

gaearon:
I think it should be safe to bring back if you do this *only* for Recompose components. Then you don't need flaky checks about whether something is a class or not—because you have full control over the components.

deepsweet:
>much better to stick with it everywhere

then as you already proposed it should be an explicit opt-in – then user is more or less aware of what's going on.

>if you do this only for Recompose components

I'm all for it if we have a proper way to do this – does Andrew's tweet mentioned above still make sense?

gaearon:
>does Andrew's tweet mentioned above still make sense?

I can't say if a tweet "makes sense". I'm not sure what you're asking. React won't do this optimization for now because it's not very useful to apply it *everywhere*. And it's something a compiler is better positioned to do.

But I'm not sure I'm seeing a problem with Recompose being a bit smarter around its own components. Maybe I'm missing something.

istarkov:
From my side I see nothing bad to accept PR which will not cause any issues we had before.

I wanna ask people how looks like your longest streak of composed recompose HOCs in current project. As having modern tools like `relay`, ready form state management libs like `Formik`, I almost lost the need to map-rename-with-and-do-something-with-Props in my projects. 

As an example my current longest recompose streak looks like below (just parsed my code base)

```
withStateHandlers
withHandlers
withPropsOnChange
withProps <- only this is functional
```

What the performance benefit will be achieved on some kind of real example? 
1% or more - hard to say without perf tests.

gaearon:
I've seen some wild `mapProps` chaining in some projects.

kof:
I landed in this discussion because I joined a project where I see an input field with 16 HOCs. I decided to dig into it because it didn't seem right. Now I am wrapping my had around why that happens and where it needs to be fixed. Right now it seems it is both - recompose and our code.

In general as an architect I am looking for a way to guide other developers on how to write code without landing in this mess of 16 HOCs input.

deepsweet:
from what I got from that tweet is that `isReferentiallyTransparentFunctionComponent` was not enough and kinda naive. I might be wrong interpreting it :)

gaearon:
We chatted with @acdlite today and agreed this might be a reasonable path:

* only "collapse" known Recompose components
* do this both in development and production

kof:
So we have a way to identify a react class based component, but we have no way to identify a functional one.

What if we could for e.g. Component constructor as a factory to create a component from a function?

```js
import {Component} from 'react'
const Button = Component((props) => <button></button>)

Button.isFunctionalComponent // true
```

Now we could do optimizations with user components as well, right

gaearon:
Why do you need to identify them? Recompose components could be tagged with its own flag which Recompose components would check.

gaearon:
Oh I missed the last sentence. I don’t think this makes a lot of sense for a few reasons. Feel free to raise a suggestion on React RFC repo but I don’t see people rewriting all components like this. In any case this seems unrelated to the discussion.

kof:
> I don’t see people rewriting all components like this. 

Me neither unless there are more reasons  to be able to detect a functional component. For now it could be just recompose.

deepsweet:
I'd vote for special static property to be able to mark any other "3rd party" HOC with it. at the same time it feels like it should be kinda internal Recompose thing exposed only for HOCs authors, at least at this point and mostly to not create unnecessary confusion.

gaearon:
I suggest not trying to over-abstract and focusing on this specific problem which should be soluble.

deepsweet:
>only "collapse" known Recompose components

>Why do you need to identify them? Recompose components could be tagged with its own flag

So are we talking about a special static property? I'd really like to allow 3rd party HOCs passed to `compose` to be "squashed" as well.

`isFunctionalHOC`?

deepsweet:
So far my idea is something like this:

```js
import { createElement } from 'react'

export const isFunctionalHocSymbol = 'IS_FUNCTIONAL_HOC_OR_BETTER_NAME'

export const createEagierElement = (type, props, ...children) => {
  if (typeof type === 'function' && type[isFunctionalHocSymbol] === true) {
    if (children.length === 0) {
      return type(props)
    }

    if (children.length === 1) {
      return type({ ...props, children: children[0] })
    }

    return type({ ...props, children })
  }

  return createElement(type, props, ...children)
}
```

Probably we can even omit `typeof type === 'function' &&` part.

PR #544, opened 2017-10-19
add optional ability for lifecycle to take fn

DarylThayil:
that also has props bound, this will allow lifecycle methods to make use
of parent props, instead of needing to rely on closures to apply things
ToDo before merging
[x] add documentation

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -524,13 +524,15 @@ Gets values from context and passes them along as props. Use along with `withCon
 
 ```js
 lifecycle(
-  spec: Object,
+  spec: Object || spec: (props: Object) => Object
 ): HigherOrderComponent
 ```
 
 A higher-order component version of [`React.Component()`](https://facebook.github.io/react/docs/react-api.html#react.component). It supports the entire `Component` API, except the `render()` method, which is implemented by default (and overridden if specified; an error will be logged to the console). You should use this helper as an escape hatch, in case you need to access component lifecycle methods.
 
 Any state changes made in a lifecycle method, by using `setState`, will be propagated to the wrapped component as props.
+
+You may also pass in a function that receives props from parent higher order components, to create the object of lifecycle methods.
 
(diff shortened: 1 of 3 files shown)


DarylThayil:
@acdlite

istarkov:
We will not accept this, please search in issues `lifecycle` `deprecate`

DarylThayil:
@istarkov  I understand the reasoning for deprecation, but for those of use using it, I simply wanted to make it more intuitive ( optionally )
If lifecycle is deprecated in the next major release, we will move off of it, but in the mean time it would be a nice add

istarkov:
Can you explain what's new gives your version of lifecycle.
I doesn't see anything what I can't do with previous version.
So even lifecycle won't be deprecated why we should accept this PR?

DarylThayil:
its true that this does not add anything you can not do, but with the current lifecycle to use props, you can not use arrow functions 
```js
lifecycle ({ 
    componentDidMount: function () { 
      console.log(this.props)
   }
})
```

vs 
```js
     lifecycle(props => ({
          componentDidMount: () => {
               console.log(props)
          }
     })
```

In addition, most recompose fns work this way, and even fns like `withProps` handle both a fn or object as a parameter.

This is more of a style nicety, and while recompose is mostly niceties that could be accomplished in other ways , this doesn't seem out of sync wth the spirit

istarkov:
Do you understand that in you realisation all lifecycle methods will receive only outdated `props` version. Any props change will not be available in them.

DarylThayil:
All of the lifecycles that rely on prop changes receive them as parameters as well correct?
your line of thinking does make sense though, since i was only using them for didMount and unMount i did not consider the impact on other lifecycle methods

istarkov:
I know what you want - you want this https://github.com/deepsweet/hocs/tree/master/packages/with-lifecycle

So please use that HOC ;-) as it's already exist

PR #551, opened 2017-11-01
Add ability to pass handler methods to all handler factory methods

trevorharwell:
This makes it easy to reference other handlers from within the body of a handler.  To achieve this functionality currently you would have to compose your handlers in the correct dependency order.  If a circular handler dependency was ever needed then that would not have previously been possible.

The one "tricky" part is what to do when there are prop-handler name collisions.  I decided that passed props should have priority over handlers when passed to the handler factory.

### DIFF
--- a/src/packages/recompose/__tests__/withHandlers-test.js
+++ b/src/packages/recompose/__tests__/withHandlers-test.js
@@ -85,7 +85,7 @@ test('withHandlers caches handlers properly', () => {
 
   handler(1)
   expect(handlerCreationSpy.callCount).toBe(1)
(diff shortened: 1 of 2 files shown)


istarkov:
Haven't understand what is this.
Examples?

If you want this https://github.com/acdlite/recompose/pull/401
We decided to not add it.

trevorharwell:
@istarkov Thanks for pointing me to #401.  That is effectively what I am looking for.  Having read through the discussion I take it this PR will be rejected as well.  It is very unfortunate that this kind of functionality is not allowed as it would be incredibly helpful.

wuct:
I'm going to close this PR, please feel free to reopen it if you have any other thought.

Issue #545, opened 2017-10-23
RFC: withStateOrUpdatedProps


tkvw:
I think it would be interesting to have a withStateOrUpdatedProps hoc. Basically it does the same as withState, but allows state to be overridden by new props. So a component can have internal state, but this can be reset. 
Currently I use it to connect redux to a component with local state.:

    import { createFactory, Component } from 'react'
    import {setDisplayName,wrapDisplayName,shallowEqual} from 'recompose';
    const withStateOrUpdatedProps = (
        stateName,
        stateUpdaterName,
        stateFromProps
    ) => BaseComponent => {
        const factory = createFactory(BaseComponent);    
        const stateFromPropsIsFunc = typeof stateFromProps === 'function';
        
        class WithStateOrUpdatedProps extends Component {
            state = {
                stateValue:
                    stateFromPropsIsFunc
                        ? stateFromProps(this.props)
                        : stateFromProps,
            };

            componentWillReceiveProps = nextProps => {
                if(stateFromPropsIsFunc){
                    const nextState= stateFromProps(nextState);
                    if(!shallowEqual(nextState,this.state.stateValue)){
                        this.setState({
                            stateValue:nextState
                        })
                    }
                }
            };

            updateStateValue = (updateFn, callback) =>
                this.setState(
                    ({ stateValue }) => ({
                        stateValue:
                            typeof updateFn === 'function' ? updateFn(stateValue) : updateFn,
                    }),
                    callback
                );

            render() {
                return factory({
                    ...this.props,
                    [stateName]: this.state.stateValue,
                    [stateUpdaterName]: this.updateStateValue,
                })
            }
        }

        if (process.env.NODE_ENV !== 'production') {
            return setDisplayName(wrapDisplayName(BaseComponent, 'withStateOrUpdatedProps'))(
                WithStateOrUpdatedProps
            )
        }
        return WithStateOrUpdatedProps
    };

Pajn:
This is something I use quite often as well. My use case local storage of quickly changed values, for example the value to a slider where the "outside" might only be updated on release or at least debounced.

istarkov:
There were a lot of similar proposals in the past, we decided to not add them. As it somehow break the idea that `view = fn(state) where fn is pure function`, 
with proposal above it will be possible to get different views on every new render, it's not that this library about. 
Also imagine you have a windowed components (in a simple form we don't update/render components which are out of current view (like lazy image loading but for rendering))
[...]

PR #555, opened 2017-11-07
Added withPropChangeHandler

laszlokardinal-dina:
I added a new helper called withPropChangeHandler, which can be used for reflecting to the changes of specified props, including mount and unmount.

My intention to add this was to eliminate the need of component classes, when we need them only for loading and unloading resources based on props (e.g. react router params). When the component does not get remounted, but only the props change I have to include the componentWillReceiveProps method too to handle the transition between pages, which adds unnecessary complexity. 

With this helper the transitions can be solved with one function which loads and unloads resources based on the presence and content of the specified props.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -37,6 +37,7 @@ const PureComponent = pure(BaseComponent)
   + [`mapProps()`](#mapprops)
   + [`withProps()`](#withprops)
   + [`withPropsOnChange()`](#withpropsonchange)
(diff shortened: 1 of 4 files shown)


istarkov:
Thank you. BTW it was better to use search before. Similar helper was proposed a lot of times in the past, and as I remember even at prev week. We decided to not accept it before and having that nothing has changed accept it now.

istarkov:
the same problems as here https://github.com/acdlite/recompose/issues/545#issuecomment-340469880

PR #560, opened 2017-11-12
Add onChange()

AlexanderShushunov:
I think it will be a very useful HOC.
In our project we have a lot of cases when we have to react on property changing. 
Examples:
- A container component has a property - *id*. It has to fetch a new document when *id* has been changed.
- We use redux-form and we have related fields. If some field has been changed, the other has to be changed according to the complicated rule.


istarkov:
Thank you, Ill close this for the same reason https://github.com/acdlite/recompose/pull/555

Issue #500, opened 2017-08-31
Pass state and props as arguments to lifecycle methods


maxhallinan:
One thing I like about recompose is that it enables me to create complex stateful components out of functions that don't depend on context. Two examples are `withState` and `withStateHandlers`. `state` and `stateUpdater` are passed as props instead of bound to `this`. This is nice because I can write more generic functions that consume these values. The same is not true for `lifecycle`. The handlers passed to `lifecycle` are given to their respective hooks without any context abstraction. So for these handlers, I either have to write that wrapper myself or write functions tied to `this`. Is there any interest in updating `lifecycle` to pass the context into the handler like `withState` and `withStateHandlers`?

deepsweet:
I think that the original idea was to give you a real part of a real class but inside of a HOC interface, with all the powerful things you can do with `this` like `this.setState()`. But for me it feels more like a low-level hacks and I personally totally agree with your points about context abstraction. That's why I created [with-lifecycle](https://github.com/deepsweet/hocs/tree/master/packages/with-lifecycle) as an alternative "trade off" which is acceptable for me.

@istarkov actually I heard such questions from my colleagues few times, like "why `lifecycle` is not following the same approach as other Recompose HOCs?" Then I found how lifecycle hooks are done in [Reassemble](https://github.com/wikiwi/reassemble#lifecycle) and was really impressed with the idea.

istarkov:
`lifecycle` is somewhere at the top place in possible deprecation list as in it's current implementation there is not a  big difference to write HOC as is
```
compose(
...
Base => class MyLifecycle extends Component {
    //  Do here anything you want much more that can be done in lifecycle
    render() { 
       return <Base {...this.props} />
    }
  }
....
)
```

There are issues with lifecycle with flowtype, and also it gives much more freedom than acceptable.
Also IMO if you really need lifecycle - may be better to write a component as is without enhancers at all.
Here https://github.com/acdlite/recompose/issues/193 author wrote:
`Recompose is a functional utility library for React components — not a state management library, or a full replacement for class components, or an entire application architecture.`

So I think we will deprecate lifecycle in nearest release.

istarkov:
Also lifecycle is a big source of issues here and this gives us +1 for deprecation ;-)

maxhallinan:
Thanks for the responses, @deepsweet and @istarkov! I'm sorry I've taken so long to respond myself. I unexpectedly found myself without internet access.

[...]

PR #564, opened 2017-11-14
add lifecycle example in docs

pokorson:
Hello,

What do you think about adding example for `lifecycle` usage. When I first used it, it wasn't clear that I should use `this` inside spec methods. 
Also is there a reason for defining `lifecycle` like this? It feels strange to use `this` in all this functional composition. I was thinking if it would be better if spec functions would get props and some `setState` equivalent method as arguments? It will increase API surface but I think it would be more consistent with other methods.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -532,6 +532,21 @@ A higher-order component version of [`React.Component()`](https://facebook.githu
 
 Any state changes made in a lifecycle method, by using `setState`, will be propagated to the wrapped component as props.
 
+Example:
+```js
+const PostsList = ({ posts }) => (
+  <ul>{posts.map(p => <li>{p.title}</li>)}</ul>
+)
+
+const PostsListWithData = lifecycle({
+  componentDidMount() {
+    fetchPosts().then(posts => {
+      this.setState({ posts });
+    })
+  }
+})(PostsList);
+```
+
 ### `toClass()`
 
 ```js


wuct:
You are right. `lifecycle` is a weird part in Recompose, which has been discussed in #500, and probably will be deprecated in the future.

pokorson:
@wuct will you consider merging just doc example?

docs/API.md
```diff
@@ -532,6 +532,21 @@ A higher-order component version of [`React.Component()`](https://facebook.githu
 
 Any state changes made in a lifecycle method, by using `setState`, will be propagated to the wrapped component as props.
 
+Example:
+```js
+const PostsList = ({posts}) => (
```
wuct: `({ posts })`

docs/API.md
```diff
@@ -532,6 +532,21 @@ A higher-order component version of [`React.Component()`](https://facebook.githu
...
+  <ul>{posts.map(p => <li>{p.title}</li>)}</ul>
+)
+
+const PostsListWithData = lifecycle({
+  componentDidMount() {
+    API.fetchPosts().then(posts => {
+      this.setState({posts: posts});
+    })
```
wuct:
```
fetchPosts().then(posts => {
  this.setState({ posts });
})
```

pokorson:
@wuct fixed

wuct:
Thank you!

Issue #407, opened 2017-06-16
using recompose with enzyme


komcal:
My enhance it's look like this
``` javascript
const enhance = compose(
  withState('labelText', 'setLabelText', ''),
  withState('hover', 'hovering', false),
  withState('labelLeft', 'setLabelLeftPosition', 0),
);
```
When I use enyzme to test component and I console.log the props like this
``` javascript
const wrapper = shallow(<Toolbox />);
console.log(wrapper.props());
```
It's show only first props `{ labelText: '', setLabelText: [Function] }`
Did you know what is happening?

istarkov:
Yes I know. Please read about what is `higher order components`.

Isaddo:
use [`mount`](https://github.com/airbnb/enzyme/blob/master/docs/api/mount.md) instead of `shallow`

komcal:
@Isaddo When I change `shallow` to `mount` I got output `{}` from console

istarkov:
Just think about this

```javascript
  const C = ({ propA, propB }) => <div>{propA + propB}</div>
  const B = ({ propA }) => <C propB={'propB'} propA={propA} />
  const A = () => <B propA={'propA'} />
  const wrapper = shallow(<A />)
```
It is almost the same what you have with enhancers. Every enhancer here - HOC creates a new Component so at top level component you have props only from top enhancer

istarkov:
To test all the props you need to use something like
```javascript
  const ToolboxClass = toClass(Toolbox) // otherwise won't work with stateless

  const Blabla = compose(
    withState('labelText', 'setLabelText', ''),
    withState('hover', 'hovering', false),
    withState('labelLeft', 'setLabelLeftPosition', 0)
  )(ToolboxClass)

  const wrapper = mount(<Blabla />)

  console.log(wrapper.find(ToolboxClass).props())
```

komcal:
I did it! Thank you so much @istarkov

komcal:
@istarkov  Have anyway to access Base Component if I import a Composed Component?

istarkov:
I always export base component, enhancer and enhanced component, 
it allows me to reuse them, and also to use code above for tests

marcusstenbeck:
Just for completeness this is what worked for me.

```js
// Toolbox.js

export const enhance = compose(
  withState('labelText', 'setLabelText', ''),
  withState('hover', 'hovering', false),
  withState('labelLeft', 'setLabelLeftPosition', 0)
);

[...]

PR #566, opened 2017-11-22
add #renameProp, #renameProps and #withReducer examples

pokorson:
@wuct another doc PR. I'm trying to introduce recompose to my team and they get a bit confuse about api description, what do you think about adding those examples?

Also it would be really nice to have some section about writing unit tests for components that uses recompose cause it gets a bit tricky. I'm still trying to wrap my head around how to write them. Should I create an issue to this discuss this further? Or is this something you discussed already?

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -194,6 +194,16 @@ renameProp(
 
 Renames a single prop.
 
+Example: 
+```js
+const ArticleComponent = ({ articleTitle }) => (
+  <div>{articleTitle}</div>
+);
+const Article = renameProp('title', 'articleTitle')(ArticleComponent);
+
+<Article title="article title" />
+```
+
 ### `renameProps()`
 
 ```js
@@ -203,6 +213,22 @@ renameProps(
 ```
 
 Renames multiple props, using a map of old prop names to new prop names.
+
+Example: 
+```js
+const ArticleComponent = ({ articleTitle, articleBody }) => (
+  <div>
+    <h2>{articleTitle}</h2>
+    <p>{articleBody}</p>
+  </div>
+);
+const Article = renameProps({
+  title: 'articleTitle',
+  body: 'articleBody'
+})(ArticleComponent);
+
+<Article title="article title" body="article body" />
+```
 
 ### `flattenProp()`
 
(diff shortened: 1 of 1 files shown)


istarkov:
I have nothing to say about reducer, but renameProps seems like useless hoc this days, Isnt it? With modern es6 syntax it can be written much more easily. Ma be we start to deprecate such hocs? 
@pokorson what are u think?

pokorson:
@istarkov I agree, personally I never used it. I'm not sure how you see it with ES6 syntax but personally I would just use #mapProps for this purpose. I guess reason for #renameProps is to slightly reduce code and better describe intention of operation you do. Not sure if it's worth to keep one more method in API for this.

and maybe do you have some comments on unit tests part? I saw you participated in few previous discussions about them?

timkindberg:
For unit testing here is what I’ve come up with: https://github.com/acdlite/recompose/issues/407#issuecomment-338309868

pokorson:
@istarkov should I remove `renameProp` and `renameProps` examples? Or can we just merge it?

istarkov:
Sorry, I don't know now, let's think more.
cc @wuct please suggest what to do

pokorson:
@istarkov sure, I also saw 2 issues regarding creating or updating documentation, will try to reach people responsible for them to get some insight of their work.

wuct:
I've never used `renameProp` and `renameProps` either. I use `mapProps` instead as well. 
I suggest we merge this PR with a deprecated warning in both the doc and the console.

pokorson:
@wuct are there any places in code I could look for such warning pattern?

wuct:
@pokorson No, there are not. We've not deprecated anything before. I've changed my mind and just adding some warning in the doc is good to me. We can write a migration guide in the Release page when we actually remove them.

docs/API.md
```diff
@@ -194,6 +194,14 @@ renameProp(
 
 Renames a single prop.
 
+NOTE: We recommend using `#mapProps` and ES6 syntax instead. This method may be deprecated in future releases.
+
+```
+renameProp('oldName', 'newName')
+->
+mapProps(props => ({...props, newName: props.oldName}))
```
istarkov: `withProps` not `mapProps`
`withProps(p => ({ newProp: p.oldProp }))`

pokorson:
@istarkov fixed

docs/API.md
```diff
@@ -194,6 +194,14 @@ renameProp(
 
 Renames a single prop.
 
+NOTE: We recommend using `#withProps` and ES6 syntax instead. This method may be deprecated in future releases.
+
+```
+renameProp('oldName', 'newName')
+->
+withProps(props => ({newName: props.oldName}))
```
wuct: We should follow the style guide. Please add spaces after `{` and before `}`.

e.g.

```js
withProps(props => ({ newName: props.oldName }))
````

pokorson:
@wuct fixed

pokorson:
hey @wuct I did `git --amend` and force push to fix your review so I guess it can't be reflected in diff right now :/ do you mind merging this one?

PR #574, opened 2017-12-16
Rethrow subscription errors in componentFromStream

### LINKED ISSUE #547: Throw errors from stream in render()?
I'm trying out `componentFromStream()` and was surprised to learn that errors are being swallowed when they bubble up to recompose because it does not subscribe to errors [here](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/componentFromStream.js#L35), only `next:` notifications.

I'd be happy to submit a pr if others think this would work:

```js

    componentWillMount() {
      // Subscribe to child prop changes so we know when to re-render
      this.subscription = this.vdom$.subscribe({
        next: vdom => {
          this.setState({ vdom })
        },
        error: thrown => {
          this.setState({ thrown })
        }
      })
      this.propsEmitter.emit(this.props)
    }

    // ...

    render() {
      if (this.state.thrown) {
        throw this.state.thrown
      } else {
        return this.state.vdom
      }
    }

```

spalger:
Fixes #547 

The `renderErrors` property is necessary for [errors emitted by unsubscription](https://github.com/acdlite/recompose/blob/f4745a0ef41d23b00e93ac2d4de7aa9ef0eef053/src/packages/recompose/__tests__/componentFromStream-test.js#L92) to be thrown.

Also includes error handling for `mapPropsStream()`

### DIFF
--- a/src/packages/recompose/__tests__/componentFromStream-test.js
+++ b/src/packages/recompose/__tests__/componentFromStream-test.js
@@ -102,3 +102,22 @@ test('completed props stream should throw an exception', () => {
 
   expect(() => wrapper.unmount()).toThrowError(/no elements in sequence/)
 })
+
+// TODO: test that this feeds into componentDidCatch() once upgraded to React 16
+test('throws stream error in render', async () => {
+  const Throws = componentFromStream(props$ =>
+    props$.map(props => {
+      if (props.throw) {
+        throw new Error('thrown from component')
+      } else {
+        return <div>okay</div>
+      }
+    })
+  )
+
+  const wrapper = mount(<Throws />)
+  expect(wrapper.text()).toBe('okay')
+  expect(() => wrapper.setProps({ throw: true })).toThrowError(
+    'thrown from component'
+  )
+})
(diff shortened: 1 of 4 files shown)


spalger:
I added error handling to `mapPropsStream()` too.

src/packages/recompose/__tests__/componentFromStream-test.js
```diff
@@ -102,3 +102,22 @@ test('completed props stream should throw an exception', () => {
 
   expect(() => wrapper.unmount()).toThrowError(/no elements in sequence/)
 })
+
+// TODO: test that this feeds into componentDidCatch() once upgraded to React 16
+test('throws stream error in render', async () => {
+  const Throws = componentFromStream(props$ =>
+    props$.map(props => {
+      if (props.throw) {
```
istarkov: Having that react have ability to intercept errors throwed from render may be we should throw error implicitly inside componentsFromProps without providing throw prop to user?

istarkov:
bad news also, looks like streaming components will be deprecated when async react will be published.

istarkov:
PS: I dont know now the reasons. So its not 100%
From my side wanna add that after 1year+ of my love with rxjs I realized that Im not smart enough to read my rxjs code written just few monthes before ;-) too hard for me.
Now I stopped using rxjs for ui ;-)

src/packages/recompose/__tests__/componentFromStream-test.js
wuct: The `throw` here is just for testing, the lib consumers won't see it ;)

Issue #563, opened 2017-11-14
[flow-typed] Update flow-typed libdef


102:
The current version of `flow-typed` in this repo is `recompose_v0.24.x.js` which is incompatible with `0.26.0`-- the current version of `recompose` itself.

Could this version be simply bumped or some changes required?

XaveScor:
fixed in https://github.com/flowtype/flow-typed/pull/1566

wuct:
@XaveScor Thank for your help! Do you mind to create another PR to update the flow-typed in this repo?

XaveScor:
@wuct Sure. I will create PR tomorrow

k:
From the flow-types repo it looks like there were no type changes, just version bumps.

TrySound:
This is not relevant. Types are published with the library.

PR #579, opened 2017-12-21
fix recompose version range for flowtype

XaveScor:
#563


istarkov:
there are also internal references into renamed folders must be changed also

TrySound:
This is not relevant

Issue #581, opened 2017-12-28
Getting an error server side rendering a withState component


abhiaiyer91:
Hello!

I have a component that uses `withState` like so:

```js
withState('item', 'setItem', ({ val }) => {
  return val;
})
```

Having upgraded to recompose 0.26.0 and React 16.2.0, i'm getting an error here 

https://github.com/acdlite/recompose/blob/master/src/packages/recompose/withState.js#L21

Where my app is destructuring stateValue, but in the case I'm experiencing `prevState` is null!

does React guarantee prevState as an object?

Any ideas?

pokorson:
@abhiaiyer91 are you sure you're passing correct props for the component? and it's not getting null for first render?

istarkov:
Real example? As `stateValue` is always defined for state https://github.com/acdlite/recompose/blob/master/src/packages/recompose/withState.js#L12-L13

panayi:
Getting the same error on server-rendering with [after.js](https://github.com/jaredpalmer/after.js)

wuct:
@panayi can you provide some example code to reproduce the issue?

paulussimanjuntak:
Hi, i'm want to ask is there a way that does not cause errors to use withState more than 1 time?
<img width="524" alt="Screen Shot 2020-06-05 at 17 28 36" src="https://user-images.githubusercontent.com/38719572/83860598-0b7b2c80-a752-11ea-8920-b609e5ea4c52.png">
and can you provide an example?
thankyou

PR #582, opened 2017-12-28
Protect the prevState variable

abhiaiyer91:
Addressing this https://github.com/acdlite/recompose/issues/581


istarkov:
Thank you. Can you provide a failing test? Then fix. Otherwise we cant accept this.

PR #584, opened 2018-01-02
use static class properties to convert toClass implicit return arrow

graingert:


istarkov:
Thank you, visually looks better. @wuct ?

wuct:
LGTM

wuct:
Thanks!

PR #595, opened 2018-01-24
Adding an optional callback argument to the `dispatch` method provided by `withReducer`

therealparmesh:
I thought of this after looking at `withState`, which provides `stateUpdater` methods that can take in callbacks. My rationale for allowing for an optional callback that receives the new state for the `dispatch` method that `withReducer` provides is that after calculating the new state, you might want to propagate state changes elsewhere. Maybe you need to call a given `onChange` prop or handle some other side effect like that.

We _could_ just execute the reducer with the old state and the action to generate the new state outside of the `dispatch` call, but I find it _really_ clunky to calculate the new state twice - once implicitly as a result of the `dispatch` call and once explicitly by executing the reducer - in order to address these types of scenarios. I am also not a fan of programmatically detecting state changes and handling the associated side effects via `lifecycle` methods.

I chose to use a plain callback over a promise solely because I did not see `recompose` making any real use of promises and I did not want to introduce a new paradigm. I am not attached to this decision.

Thank you for this amazing library! Please let me know what you think of this idea.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -336,7 +336,13 @@ withReducer<S, A>(
 
 Similar to `withState()`, but state updates are applied using a reducer function. A reducer is a function that receives a state and an action, and returns a new state.
 
-Passes two additional props to the base component: a state value, and a dispatch method. The dispatch method sends an action to the reducer, and the new state is applied.
+Passes two additional props to the base component: a state value, and a dispatch method. The dispatch method has the following signature:
+
+```js
+dispatch(action: Object, ?callback: Function): void
+```
+
+It sends an action to the reducer, after which the new state is applied. It also accepts an optional second parameter, a callback function with the new state as its only argument.
 
 ### `branch()`
 
--- a/src/packages/recompose/__tests__/withReducer-test.js
+++ b/src/packages/recompose/__tests__/withReducer-test.js
@@ -67,3 +67,30 @@ test('receives state from reducer when initialState is not provided', () => {
 
   expect(component.lastCall.args[0].counter).toBe(0)
 })
+
+test('calls the given callback with new state after a dispatch call', () => {
+  const component = sinon.spy(() => null)
+  component.displayName = 'component'
+
+  const initialState = { counter: 0 }
+
+  const reducer = (state, action) =>
+    action.type === SET_COUNTER ? { counter: action.payload } : state
+
+  const Counter = compose(
+    withReducer('state', 'dispatch', reducer, initialState),
+    flattenProp('state')
+  )(component)
+
+  mount(<Counter />)
+  const dispatch = sinon.spy(component.firstCall.args[0].dispatch)
+  const callback = sinon.spy()
+
+  dispatch({ type: SET_COUNTER, payload: 11 }, callback)
+  expect(dispatch.calledBefore(callback)).toBe(true)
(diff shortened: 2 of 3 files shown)


istarkov:
Hi, Thank you.
It was created as some kind of simplified redux, no callback there https://redux.js.org/docs/api/Store.html#dispatch
We have no analog of middlewares to call side effects on action complete etc.
I personally dont like the idea to provide api extension the only use case it has is to call some side-effect and somehow broke similarity to redux dispatch. In all such cases you can call sideeffect in handler with just action (and compute state where it needed using the same logic as in store)

ShibumiParmesh:
Thank you for the feedback!

While I definitely see that this is a simple `redux` analog, I still think that something that models this functionality of [`dispatch`](https://redux.js.org/docs/api/Store.html#dispatch) could really benefit `recompose`.

> The store's reducing function will be called with the current getState() result and the given action synchronously. Its return value will be considered the next state. It will be returned from getState() from now on, and the change listeners will immediately be notified.

istarkov:
I havent used withReducer ;-) @wuct  what are you think about, may be in absence of middlewares this is not so bad as I wrote?

therealparmesh:
It'd be great to have middlewares like in `redux` to manage side effects cleanly. 😄But I imagine that adding support for middleware would be tough, given that the `withReducer` HOC is basically emulating the `redux` `store` with the main difference being that the HOC is bound to React's API internally.

There are already some differences in the `dispatch` implementations because of this. The `dispatch` from `withReducer` is technically asynchronous since it relies on and returns a `setState` call internally. So, it does not return the same thing as `dispatch` from `redux` as is.

So I ask - that in lieu of middleware support and that given that the `dispatch` implementations are seemingly going to be slightly different - whether or not an API extension like this is an acceptable way to address simple side effect management when using the `withReducer` HOC.

azz0r:
My suggestion would be to pass in a callback to the payload, e.g.

https://github.com/azz0r/fedsimulator/blob/ppvs/src/actions/data.js#L9
https://github.com/azz0r/fedsimulator/blob/ppvs/src/reducers/roster.js#L24

But I believe this is considered bad practise, you'd be expected to use componentWillReceiveProps to check the nextProps change then trigger your "callback".

wuct:
@therealparmesh @istarkov I don't like the idea to extend API surface for side effects, too. However, as @therealparmesh said, we've already provided this kind of surface in `withState`, and if we consider `withReducer` as another `setState` wrapper, it makes sense that we create the same surface for it. Moreover, I don't think we will support for middlewares, so it's okay for me to have this callback.

@azz0r IMHO it's not a good idea because impure reducers are harder to test and trace code.

therealparmesh:
@wuct @istarkov Great! Thank you for taking the time to look at this. What should the next steps be?

wuct:
Looks like it is good to be shipped. @istarkov any thought?

tlginn:
Any update on if/when this will be released? It would be helpful in a project where I'm currently using recompose. 

There are workarounds for sure, but I'm not a fan of having an impure reducer either and `componentWillReceiveProps` is distant from the code that is dispatching the action. Having this would be a great convenience.

istarkov:
Oops sorry,
@wuct I see nothing bad, please merge.

wuct:
Thank you @therealparmesh

PR #604, opened 2018-02-06
Context in withContext ;)

Anmo:
Use `call` to bind the `getChildContext` to the `Component`.

I leave the `this.props` to maintain compatibility. Because we are now (binded), we don't need to pass the `props`.
If we don't use the `this.props` anymore, we could do it like this: `getChildContext = getChildContext.bind(this)` instead.


istarkov:
Hi
Just one question Why?
What problem does it solves?

istarkov:
1) Using `this.props` in getChildContext function is not about this library is about ;-) we are trying to use pure functions anywhere.

2) What if you pass an anonymous function `() => { blablabla }` as `getChildContext` it will not receive your `this` as I know and will have it's own `this` ;-)

I'll close this,
In any case thank you!!!, 
there are soon be a lot of opportunities to improve this library having that React will start to change.

Issue #614, opened 2018-02-27
Performance optimizations have been removed


Siegrift:
There is a link to watch Andrew's talk on Recompose at React Europe (https://www.youtube.com/watch?v=zD_judE-bXk). However, performance optimizations he speaks about have been removed in #538.  

Maybe adding some info about this change under the link would help. 

also the [link](https://github.com/acdlite/recompose/blob/master/docs/performance.md) explaining the performance is outdated.

wuct:
@Siegrift Thanks for bringing this up! Would you like to edit it?

Siegrift:
I can try

wuct:
Closed by #615

PR #615, opened 2018-02-28
Update docs & readme

Siegrift:
I am not sure if we want to remove `performance.md` completely or just the outdated part.

Fixing #614


wuct:
Thanks!

PR #621, opened 2018-03-04
Webpack 4

### LINKED ISSUE #619: Use webpack's side-effects flag to enable tree shaking in Webpack 4
This library is missing the "sideEffects" : false in the package.json to enable Webpack 4 tree shaking.

Lodash-es example:
https://github.com/lodash/lodash/blob/master/package.json#L10

This would be a (better) alternative for:
https://github.com/acdlite/recompose#using-babel-plugin-lodash

wuct:
Close #619


TrySound:
I think this is not relevant.

https://github.com/acdlite/recompose/blob/master/.size-snapshot.json#L16-L19

PR #623, opened 2018-03-06
 DOC API compose section recomposing 

axelmm:
grrr, @#$%^  **you're right**  -  I wrongly readed intention - I assumed these are usage examples (I saw many hoc1(hoc2(... used directly on component in export default statement) while they're definitions of HOCs ... prepared for usage/reusing    ..... 

I was watching and not seeing ... that BaseComponent is only parameter definition in ES6 function definition - 1-st one doesn't need this explicitely
I see it **a bit missleading**, now.  In all other code in this section BaseComponent is a real component to be enhanced - this makes some common context while these lines are out of it.

Maybe making it a little more verbose  (updated file) ?

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -9,10 +9,14 @@ const EnhancedComponent = hoc(BaseComponent)
 This form makes HOCs (sometimes called **enhancers**) composable:
 
 ```js
-const composedHoc = compose(hoc1, hoc2, hoc3)(BaseComponent)
+// instead of defining usual HOC (ES6 function) definition
+const composedHoc = Component => hoc1(hoc2(hoc3(Component)))
 
-// Same as
-const composedHoc = BaseComponent => hoc1(hoc2(hoc3(BaseComponent)))
+// we can express it shorter
+const composedHoc = compose(hoc1, hoc2, hoc3)
+
+// In both cases we can use this HOC in the same way
+const EnhancedComponent = composedHoc(BaseComponent)
 ```
 
 Most Recompose helpers are **functions that return higher-order components**:
--- a/docs/API.md
+++ b/docs/API.md
@@ -9,7 +9,7 @@ const EnhancedComponent = hoc(BaseComponent)
 This form makes HOCs (sometimes called **enhancers**) composable:
 
 ```js
-const composedHoc = compose(hoc1, hoc2, hoc3)
+const composedHoc = compose(hoc1, hoc2, hoc3)(BaseComponent)
 
 // Same as
 const composedHoc = BaseComponent => hoc1(hoc2(hoc3(BaseComponent)))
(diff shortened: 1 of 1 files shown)


wuct:
We use `BaseComponent` to name the last argument of all HOCs in the source code, for example [here](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/withStateHandlers.js#L7) and [here](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/pure.js#L6), so I believe using the same word in the doc is better.

axelmm:
This is why I wrote 'a bit missleading' - I know it's good, now ... but lacking of/mixed context leaded me wrong way. This is very introductive section where reader's context is based on earlier experience ... where probably wasn't needed separated HOC definition.

Adding before all other lines 
```js
const BaseComponent = props => {...}
```
we'd have real `BaseComponent` followed with `BaseComponent` (not related to earlier statement!) only as parameter and again real `BaseComponent` enhanced by `composedHoc`. The same name, mixed context.

Consistency in code is ok, but in examples it can be missleading. If `BaseComponent` is used in definitions than use AnotherComponent **in usage examples** ... just to **minimize missunderstanding**.
```js
// instead of defining HOC as usual (ES6 function)
const composedHoc = BaseComponent => hoc1(hoc2(hoc3(BaseComponent)))

// we can express it shorter
const composedHoc = compose(hoc1, hoc2, hoc3)

// In both cases we can use this HOC in the same way
const EnhancedComponent = composedHoc(LayoutComponent)
```

wuct:
We don't use `Component` here because it might be confused with `React.Component`. `LayoutComponent` is coming from nowhere, so it might be confused. too. How about

```js
// instead of defining HOC as usual (ES6 function)
const composedHoc = BaseComponent => hoc1(hoc2(hoc3(BaseComponent)))

// we can express it shorter
const composedHoc = compose(hoc1, hoc2, hoc3)

// In both cases we can use this HOC in the same way
const EnhancedComponent = composedHoc(BaseComponent)
```

axelmm:
Could be .. but still mixing context - maybe explain this directly, more complete example?
```js
// component to be wrapped
const BaseComponent = props => {...}

// instead of defining HOC as usual (ES6 function)
const composedHoc = BaseComponent => hoc1(hoc2(hoc3(BaseComponent)))

// we can express it shorter
const composedHoc = compose(hoc1, hoc2, hoc3)

// In both cases we can use this HOC in the same way
const EnhancedComponent = composedHoc(BaseComponent)
```

> **By convention** we're using `BaseComponent` almost everywhere as function parameter.
This way `BaseComponent` in second statement **is not related** to component defined earlier (and used at the end) - `composedHoc` (as a function) can be used to enhance any other component (passed argument).

istarkov:
IMO compose is one of the first function - developers meet with starting learning functional ideas. No need to extend somehow?

istarkov:
I'll close this for now.

PR #627, opened 2018-03-15
Object Rest/Spread compliance

henryqdineen:
This is an error I ran into while statically analyzing this code. 

In newer versions of `babylon` it will error if the rest param is not the end when desctructuring.

See here: https://github.com/babel/babylon/blob/master/CHANGELOG.md#eyeglasses-spec-compliancy-6


wuct:
Thanks!

PR #628, opened 2018-03-15
Simplify compose

crimx:
Hi! A bit of refactoring. Since function can only return single value, there is no need to create wrapper functions for every compose item.

### DIFF
--- a/src/packages/recompose/compose.js
+++ b/src/packages/recompose/compose.js
@@ -7,5 +7,11 @@ export default function compose(...funcs) {
     return funcs[0]
   }
 
-  return funcs.reduce((a, b) => (...args) => a(b(...args)))
+  const innerFunc = funcs[funcs.length - 1]
+  const restFuncs = funcs.slice(0, -1)
+  const reducer = (result, fn) => fn(result)
+
+  return function composed(...args) {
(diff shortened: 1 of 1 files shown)


wuct:
Actually, we used to have the similar version of `compose` before [this commit](https://github.com/crimx/recompose/commit/1c3d2b38edb394db1144af9f078a73008c612f7f#diff-8670130e933c8326c511499187dad579).

wuct:
Despite losing some performance, the current one is more readable.

crimx:
I've read the initial commit before I started this pull request. And I realized the current implementation was for readability.

I think my PR satisfies both sides. It's actually more readable IMHO because it follows the underlying direction of compose.

crimx:
What I meant was it reflects the idea that `compose(a, b, c)(...) === c(b(a(...)))` better than using HOCs.

wuct:
Moreover, the current one is same as [the redux one](https://github.com/reactjs/redux/blob/b4fb08133c95094a4b293a9ab434d1d5dd657527/src/compose.js#L21), so it might be more familiar to the React community 😄

crimx:
Understood. That makes sense.

PR #630, opened 2018-03-20
Implement withPromiseHandler

MrBr:
Hey,

Just want to share an idea here, I think it fits recompose.

Basically, when working with promises in a component I want to preserve React lifecycle or to be more precise be notified that promise is resolved through props update. There is also issue with loading status, there is no easy way to tell if a promise is resolved (usually it ends being saved in the component state).

`withPromiseHandler` provides a method to the wrapped component for registering a promise and in return, it will handle promise. When the promise is resolved, successfully or with error, the result will be saved in the state of the HOC and passed to the wrapped component. 

A bonus is `loading` status which is added to the promise object so that the wrapped component can easily and without any additional state handle that case.

The ultimate thing is that every time promise changes, a wrapped component gets a chance to render that new state automatically.

If you think this fits recompose I can do any needed changes and write tests.

Cheers!

### DIFF
--- a/src/packages/recompose/withPromiseHandler.js
+++ b/src/packages/recompose/withPromiseHandler.js
@@ -1,0 +1,59 @@
+import React, {Component} from 'react';
+
+const DEFAULT_PROMISE_NAME = 'response';
+
+const withPromiseHandler = (WrappedComponent, promiseNames = [DEFAULT_PROMISE_NAME]) => {
+  return class extends Component {
+    constructor(props) {
+      super(props);
+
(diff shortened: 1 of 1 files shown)


istarkov:
Hey, we prefer to not add so specialised components especially with promises and other async stuff ;-) Also wanna add that realisation above lack of details like canceling currently executed promises to avoid setState on unmounted components etc.
Also just in few months we will get React suspense api https://reactjs.org/blog/2018/03/01/sneak-peek-beyond-react-16.html so most of such things will become obsolete IMO

PR #632, opened 2018-03-26
Pass blacklist argument to hoist-non-react-statics using hoistStatics

pex:
[hoist-non-react-statics](https://github.com/mridgway/hoist-non-react-statics) supports an [undocumented argument `blacklist`](https://github.com/mridgway/hoist-non-react-statics/blob/master/src/index.js#L34) in order to blacklist static properties that shall not be copied.

This PR adds that option to `recompose/hoistStatics`.

### DIFF
--- a/src/packages/recompose/__tests__/hoistStatics-test.js
+++ b/src/packages/recompose/__tests__/hoistStatics-test.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { createFactory } from 'react'
 import { mount } from 'enzyme'
 import sinon from 'sinon'
 import { hoistStatics, mapProps } from '../'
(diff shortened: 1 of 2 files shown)


istarkov:
Just curious is it possible to use hoist-non-react-statics instead of current implementation, may be we should deprecate current and recommend use 3rd party lib instead of trying to make current implementation similar to it. As it do one thing and do it well?

istarkov:
oops we already use it ;-)

istarkov:
@wuct I see nothing bad here

wuct:
LGTM. Thank you!

wuct:
@pex Can you kindly update the API doc as well?

PR #634, opened 2018-03-27
docs: Remove misleading optional annotation

FezVrasta:
According to the docs, `EnhancedComponentProps` is the type of the props that will be assigned to the generated component.

If this is the case, `text` is never going to be `undefined` or `null` because our HOC is making sure to set it to `string` in any possible case.

### DIFF
--- a/types/README.md
+++ b/types/README.md
@@ -11,7 +11,7 @@ Example:
 import type { HOC } from 'recompose';
 
 type EnhancedComponentProps = {
-  text?: string,
+  text: string,
 };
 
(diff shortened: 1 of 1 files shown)


istarkov:
Enhanced component can can define text or not. Base component will have text defined always.

All is right.

FezVrasta:
So "EnhancedProps" are going to be the outer props of the enhanced competent? I think the wording could help some improvement in this case...

istarkov:
Just props of enhanced component, components have no outer or inner props. Wording is hard ;-)

PR #636, opened 2018-03-28
RFC: Add a HOC that prints the props of a component.

jstralko:
This is helpful when debugging or looking at props in a presentational
component.


istarkov:
Thanks, BTW it was already discussed in issues. There is no need in creating something what can be easily achieved in few symbols using current api
i.e. `withProps(console.log)`

jstralko:
awesome! I wasn't aware of it, but yes that accomplishes what I need...

wuct:
It’s described in the README
Gerb Stralko <notifications@github.com>於 2018年3月29日 週四，上午10:56寫道：

PR #643, opened 2018-04-09
better compose

zeakd:
Hi. it's not a issue, but made simple compose without 'if'. thanks.

### DIFF
--- a/src/packages/recompose/__tests__/compose-test.js
+++ b/src/packages/recompose/__tests__/compose-test.js
@@ -21,6 +21,6 @@ test('compose returns the identity function if given no arguments', () => {
 })
 
 test('compose returns the first function if given only one', () => {
-  const fn = () => {}
-  expect(compose(fn)).toBe(fn)
+  const fn = x => x * x
+  expect(compose(fn)(3)).toBe(fn(3))
 })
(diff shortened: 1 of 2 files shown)


istarkov:
I like such changes, prev realusation of compose usually had same behaviour as here https://github.com/reactjs/redux/blob/master/src/compose.js
I think because of perf reasons for some strange use cases (no need in compose for 1 function or 0 at all) so I think we can accept this cc @wuct your thoughts?

ryota-murakami:
interesting change, and i wonder how much performance increase on today's CPU, that has heighly optimized *if else branch prediction*.
but i'm not familiar V8 engine, so you make me want to know V8 more deeply!

about @istarkov mentioned some strange use cases, i came across through contribute to [spectrum](https://spectrum.chat/)).
you could see like `export const Card = compose()(CardPure);` fix temporarily code:smile:
https://github.com/withspectrum/spectrum/commit/a245109a5004d3d07f4f966a418354118e210e0f

(cc @mxstbr

wuct:
Yes, I think we can accept it. The new approach is the reason why the "identity" function has the name "identity," and I like it. :)

dokai:
Hi! I came upon this PR while reviewing the proposed upgrades to our app's dependencies.

I appreciate the slightly shorter version of the implementation but had one question come up. It seems the initial motivation was to optimize away the `if` check and therefore do less work.

However, it seems that this trades a one-time cost at composition time `(*)` to a run-time cost on each invocation of the composed function due to adding the identity function in the composition chain unconditionally.

I'm not familiar with the details on how the engines would  actually treat or optimize this but in general the cost of a function call is usually higher than an `if` check.

`(*)` I'm assuming that the common pattern is to bind the composed function to a variable that is used to invoke the function therefore having to evaluate the `if` only once, e.g.
```javascript
const myFunc = compose(fn1, fn2, f3);
myFunc('foobar')
```

PR #645, opened 2018-04-09
Refactor rollup config

TrySound:
- move all bundles into dist folder
- add size snapshot
- remove treeshake test


istarkov:
Thank you!!!

PR #646, opened 2018-04-12
Update dependencies before changes according to React 16.3+

istarkov:
In order to provide changes came in React 16.3 here is dependencies and tests update.
To simplify next commits.


package.json
```diff
@@ -90,5 +88,8 @@
   "devEngines": {
     "node": "5.x",
     "npm": "3.x"
+  },
+  "dependencies": {
+    "enzyme-adapter-react-16": "^1.1.1"
```
wbyoung: I know this was already merged, but shouldn't this be in `devDependencies`?

PR #647, opened 2018-04-13
New lifecycles support

istarkov:
In async mode `propsMapper` in `withPropsOnChange` can be called multiple times
it can be avoided using memoize see this commit https://github.com/acdlite/recompose/pull/647/commits/83fc7cadb0af2ef29fd3abbb0a37e634ab5ceddf
but it does not solve some edge cases as memoize will work on all `props` so multiple calls will still be possible. So I think to remove it.


istarkov:
next steps
- New context api for existing HOCs, 
- new HOCs for new Context api,
- move streaming HOCs into other package.

PR #649, opened 2018-04-18
Include flow types into library

istarkov:
This removes the need of flow-typed, manual copying anything etc
Now types are provided with recompose


istarkov:
@TrySound can you help me a little, do I need to place index.js.flow into dist folder too, or placing it as it is will be enough

package.json
```diff
@@ -10,13 +10,14 @@
...
-    "test": "jest",
+    "test": "jest && flow",
     "test:watch": "cross-env BABEL_ENV=cjs jest --watch",
     "release": "node scripts/release.js",
     "postinstall": "node scripts/installNestedPackageDeps.js",
     "format": "prettier --semi false --trailing-comma es5 --single-quote --write 'scripts/*.js' 'src/packages/*/*.js' 'src/packages/*/!(node_modules)/**/*.js'",
     "precommit": "lint-staged",
-    "prepush": "yarn test"
+    "prepush": "yarn test",
+    "flow": "flow"
```
TrySound: it's not necessary with yarn

PR #660, opened 2018-05-01
 Update rxjsObservableConfig for rxjs 6

wuct:
Need some discussion before merging:

1. There were `rxjs4ObservableConfig.js` and `rxjsObservableConfig.js` (for rxjs 5) before. This PR upgrade `rxjsObservableConfig.js` directly to adopt rxjs 6, but it's a breaking change. We might want to create separate files for both rxjs 5 and rxjs 6.

2. Rxjs used to polyfill `Symbol.observable`, but it has been removed since https://github.com/ReactiveX/rxjs/pull/3387. We might need to add a notice about polyfill `Symbol.observable` to the doc.

### DIFF
--- a/src/packages/recompose/__tests__/componentFromStream-test.js
+++ b/src/packages/recompose/__tests__/componentFromStream-test.js
@@ -1,7 +1,15 @@
 import React from 'react'
 import { mount } from 'enzyme'
 import { Observable, Subject } from 'rxjs'
-import sinon from 'sinon'
+import {
+  first,
+  last,
+  startWith,
+  map,
+  mapTo,
+  combineLatest,
+  tap,
+} from 'rxjs/operators'
 import rxjsConfig from '../rxjsObservableConfig'
 import { componentFromStreamWithConfig } from '../componentFromStream'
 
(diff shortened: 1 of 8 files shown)


src/packages/recompose/__tests__/componentFromStream-test.js
```diff
@@ -93,20 +109,3 @@ test('complete props stream before unmounting', () => {
   wrapper.unmount()
   expect(counter).toBe(0)
 })
-
-test('completed props stream should throw an exception', () => {
```
wuct: It seems like rxjs 6 doesn't throw this error anymore.

istarkov:
What are u think to split library on 2, ie recompose and recompose-rx(-observable, -blabla) and move all the streaming parts out of the recompose library. 

Also I think we can move on https://github.com/facebook/react/tree/master/packages/create-subscription to support async in the future

wuct:
It would be a good idea to separate them. But I'm thinking about keeping the es-observable parts remain in Recompose and move only those 3rd-party-config files into other libraries. The reason for me to separate them is to have better version controlling.

And yes, I think we need to introduce create-subscription to handle streaming events.

OliverJAsh:
RxJS 6 has been out for awhile now. Is this PR good to merge? Re. the suggestion of moving 3rd-party-config files into other libraries, perhaps keep that idea to a separate PR, so we can unblock this one?

elmarburke:
For those people waiting for this PR, there is a simple configuration possible to get recompose working with RxJS@^6.0.0:

```js
import { from } from 'rxjs';

setObservableConfig({
  fromESObservable: from,
});
```

PR #663, opened 2018-05-04
Add withRenderProps

pomber:
From this [tweet](https://twitter.com/icelabaratory/status/992123859590631424). Let me know if there's anything missing.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -56,6 +56,7 @@ const PureComponent = pure(BaseComponent)
   + [`getContext()`](#getcontext)
   + [`lifecycle()`](#lifecycle)
   + [`toClass()`](#toclass)
+  + [`withRenderProps()`](#withRenderProps)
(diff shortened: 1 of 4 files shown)


istarkov:
Super! Thank you, I will check at sunday|monday, seems like travis is broken now.

wuct:
Is TravisCI broken now?

istarkov:
Thanks, Ill add flowtypings and make release.

sylvain-hamel:
What are the use cases and benefits of withRenderProps? I asked [this question on SO](https://stackoverflow.com/questions/50933396/what-are-the-use-cases-and-benefits-of-withrenderprops)  but did not get any answer. I understand the render prop pattern. What I'd like to know is when/why I would convert some of recompose's HOCs (or my own HOCs) this way, using withRenderProps.

PR #665, opened 2018-05-08
Fix order of parameters of `_.omit()` in docs for `mapProps()`

michalczaplinski:
For reference: https://lodash.com/docs/4.17.10#omit


istarkov:
lodash/fp

Issue #666, opened 2018-05-10
withPropsOnChange: when shouldMap returns false, subsequent calls have outdated "prevProps"


grrowl:
When using withPropsOnChange with "shouldMapOrKeys" function first argument (a.k.a.`shouldMap`, signature `(prevProps, nextProps): Boolean`), prevProps is only updated when the previous call to `shouldMap` returned `true`. 

I believe this is incorrect — `shouldMap` signals whether `computedProps` should be updated, and `prevProps` should _always_ be updated, ready for the next call to `shouldMap`/

### Repro steps

1. On first change, `shouldMap` is called with `prevState.prevProps`
2. Because this is within `getDerivedStateFromProps`, if `shouldMap` returns false, it returns null (does not update state)
3. Therefore the next call to `shouldMap`, it is passed not the "current" props but the original, on-mount props.
  * Or, after some changes, "prevProps" will actually be "props from the last time `shouldMap` returned `true`"

I have pushed PR #667 with this change :)

istarkov:
see PR comment

motiazu:
+1 on this. This change broke old behavior and I heavily rely on `withPropsOnChange`. But now `prevProps` are not reliable in certain cases. And this is causing a lot of flack from the anti recompose guys here also :(

motiazu:
It took me a day to realize the reason my `oldProps` are not up to date because `withPropsOnChange` changed it's behavior a bit and release notes didn't mention that. Release notes should state that behavior was changed.

But I think something might have been missed here. The problem not only happens if you "calculate something based on some propsA, and don't use that propsA in decision to recalculate".
I have a case with something like this:
```
withPropsOnChange(
  ({ auth }, { auth: nextAuth }) => !auth.hasError && nextAuth.hasError,
  ({ auth, onExpectationFailed }) => {
    if (_.get(auth, 'error') === httpStatus.EXPECTATION_FAILED) {
      onExpectationFailed();
    }
  }
),
```
Because my `shouldMap` returns false when transitioning from hasError=true to false, the next time I have an error my old props indicate hasError=true even though they are false. So if I'm looking inside a prop's fields for my comparison, this change breaks my code.
If you would keep the oldProps in derived state, it will be consistent with previous implementation.

istarkov:
I don't want to be consistent with previous, IMO previous implementation was wrong.
I wrote why I think so see links (PR, and link iside PR)
https://github.com/acdlite/recompose/pull/667
[...]

Issue #651, opened 2018-04-20
Bug: `WithPropsOnChange` accesses `prevProps` on `prevState`


tdeekens:
Thanks for the great release of 0.0.27!

However, with the recent release of `0.0.27` we started using the improved lifecycle methods of React. I noticed a possible regression in introducing them in `withPropsOnChange`. 

https://github.com/acdlite/recompose/blob/936e49fbfd3a70f1beedc2578c39dbdabcf8ef99/src/packages/recompose/withPropsOnChange.js#L25-L30

I wonder how `prevProps` can exist on `prevState`. Furthermore, how it's meant to work in conjunction with the `propsMapper`

https://github.com/acdlite/recompose/blob/936e49fbfd3a70f1beedc2578c39dbdabcf8ef99/src/packages/recompose/withPropsOnChange.js#L13-L17

`getDerivedStateFromProps` will not have access to `this` in that it's a `static` method. As a result we should not be able to access the `prevProps` at all.

istarkov:
About what bug are you talking? Can you reproduce it? Be sure we have tests on every enhancer in the library.

So every time this line https://github.com/acdlite/recompose/blob/936e49fbfd3a70f1beedc2578c39dbdabcf8ef99/src/packages/recompose/withPropsOnChange.js#L26 is true we update `prevProps` in a state.

And I know that it's slightly different from previous behaviour where we used real latest prepProps but I think that no one will be affected. If there will be issues I'll add `return {prevProps: nextProps}` into this line https://github.com/acdlite/recompose/blob/936e49fbfd3a70f1beedc2578c39dbdabcf8ef99/src/packages/recompose/withPropsOnChange.js#L33

As an example what worked before and will work not same now.

```
withPropsOnChange(({a}, {a: b} ) =>  a - b > 1, () => console.log(1));
```

before, if a increased by +1 each time, this will not be called on update
now it will be called on second update.

If this somehow affects you I'll change this to previous behaviour. 

BTW current looks more right to me - as we recalculating something only if state of previous calculation has changed.

tdeekens:
Sorry for the too vague description of a potential but. The facts that prevProps is accessed from prevState looked stange to me. It‘s where our app „crashes“ with an exception that prevProps does not exist on prevState. 

I will have a more indepth look at both recompose and our application code. I will try to deduct a mininal repro case in eg a CodeSandbox. 

Thanks a lot for the quick response.

tdeekens:
[...]

PR #667, opened 2018-05-10
withPropsOnChange: always update state.prevProps

grrowl:
...even if shouldMap returns false (we won't update computedProps)

Fixes acdlite/recompose#666

### DIFF
--- a/src/packages/recompose/__tests__/withPropsOnChange-test.js
+++ b/src/packages/recompose/__tests__/withPropsOnChange-test.js
@@ -43,3 +43,63 @@ test('withPropsOnChange maps subset of owner props to child props', () => {
   expect(component.calledThrice).toBe(true)
   expect(mapSpy.callCount).toBe(2)
 })
+
+test('withPropsOnChange maps subset of owner props to child props with shouldMap function', () => {
+  const component = sinon.spy(() => null)
+  component.displayName = 'component'
+
+  const mapSpy = sinon.spy()
+  const StringConcat = compose(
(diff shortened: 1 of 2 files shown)


istarkov:
Im especially did that change, and compare with props used to calculate result.
As it looks more right for me. (as like as memoize where props used to be memoize key)
What is usecase you need behaviour above?

istarkov:
https://github.com/acdlite/recompose/issues/651#issuecomment-383163549

grrowl:
I pushed a test which demonstrates the issue. Passes on `v0.26.0` and fails on `master`.

Our use case is combining multiple request statuses (loading, error, fulfilled) into more relevant component props (readOnly, loading, error). When the "previous state" became inconsistent, we stopped transition state correctly.

If this is a deliberate API change we're happy to update our use, but it should be listed as a breaking change and the docs updated. FWIW I think it's more intuitive to consistently be passed the previous props

istarkov:
Yep I forgot to add this behaviour into releases section.
The idea was to make behaviour of withPropsOnChange to be more near to behaviour of memoize function. As it looks strange to calculate something based on some `propsA`, and don't use that `propsA` in decision to recalculate. As like as example I provided in comment link above - I will expect that something will change if delta>1 but with previous behaviour result will never change.

There is the other problem will be in near future, the component will not work as expected with async behaviour and possibly would need to be rewritten to use memoize internally, and changes above will help to achieve this a little (or may be not ;-)).

PR #673, opened 2018-05-18
Refactor deprecated lifecycle hooks in `componentFromStream`

jonasws:
The usage of `componentWillMount` and `componentWillReceiveProps` is deprecated as of React v16. This PR moves the event emissionnn upon the component being rendered with new props into other lifecycle hooks.

Even though the tests are still running fine here, there may be some performance hitch that have went by me, so feel free to comment!

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,18 +1,18 @@
 {
   "lib/packages/recompose/dist/Recompose.umd.js": {
-    "bundled": 85035,
-    "minified": 31377,
-    "gzipped": 10006
+    "bundled": 84914,
+    "minified": 31331,
+    "gzipped": 10001
   },
   "lib/packages/recompose/dist/Recompose.min.js": {
-    "bundled": 81529,
-    "minified": 30125,
-    "gzipped": 9618
+    "bundled": 81408,
+    "minified": 30079,
+    "gzipped": 9615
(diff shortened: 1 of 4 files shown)


istarkov:
Hey thank you, this will not solve the near problems with the future react version.
We need to rewrite componentFromStream using this package https://github.com/facebook/react/tree/master/packages/create-subscription

istarkov:
If you are interested in, please welcome.

jonasws:
I can’t make any promises, but I will definitely look into it!

jonasws:
@istarkov Do we want to still merge this, or at least rename the lifecycles to use the `UNSAFE` prefix?

istarkov:
Changing `componentWillMount` -> `componentDidMount` is a serious breaking change as you break the data for the first render and break ssr rendering too, so even it's not in tests it can cause a lot of problems. 
In such case I prefer to make just one switch on a `create-subscription` solution
Having that 
- the only need of change above is to support async react.
- minimize breaking changes

jonasws:
@istarkov I have made an initial attempt on implementing this HOC using `create-subscription`. Howver, my experience with the `enzyme` library is a bit limited here: I can't seem to access the node rendered by the `Subscription` render prop. Any pointers here?

PR #686, opened 2018-06-08
Don't throw if pick gets undefined

bkniffler:


istarkov:
see comment in corresponding issue.

Issue #693, opened 2018-06-25
Getting all render arguments in fromRenderProps


itsacezon:
Apollo React's `<Mutation />` component passes multiple arguments in its render prop function (https://www.apollographql.com/docs/react/essentials/mutations.html#basic). This means that in `fromRenderProps` only the mutator is passed and not the remaining arguments.

Maybe we can refactor this line (https://github.com/acdlite/recompose/pull/677/files#diff-73d23a5af881a3abeef802a6a9db71f6R15) and spread the arguments when passing to `propsMapper`?

```javascript
...
[renderPropName]: (...props) =>
  baseFactory({ ...ownerProps, ...propsMapper(...props) }),
...
```

evenchange4:
@itsacezon Thanks for the feedback. :) 

I created a PR #694 which supports for multiple arguments in `fromRenderProps` HOC. 
Would you like to share more detail about your use case for using Apollo with `fromRenderProps`?

itsacezon:
Thanks for the quick response! 🎉

I'm making a generic HOC that returns a decorated `mutate` function that passes the `variables` object to the GraphQL query. When I saw the `fromRenderProps` update, it made my life easier!

Here's an excerpt; my goal is to reduce boilerplate code for GraphQL mutations:

```javascript
function withMutation (mutation, config = {}, mutateOptions = {}) {
  const handlerName = config.name || 'mutate';
  const _Mutation = withProps({ mutation, ...config })(Mutation); // from `apollo-react`

  return compose(
    fromRenderProps(_Mutation, (mutate, result) => ({ mutate, result }),

    withHandlers({
      [handlerName]: ({ mutate, ...rest }) => params => {
        // So options like `optimisticResponse` and `variables` can use the parameters passed
        // when calling the mutator, as well as the owner props (including the `result` object)
        const options = typeof mutateOptions === 'function'
          ? mutateOptions(params, rest)
          : mutateOptions;

        mutate(options);
      },
    })
  );
}
```

PR #694, opened 2018-06-25
feat(fromRenderProps): add support for multiple arguments (Apollo usecase) #693

evenchange4:
Add support for multiple arguments in `fromRenderProps` HOC.

for issue #693


istarkov:
Thank you! Good catch!

PR #707, opened 2018-07-23
Fix withPropsOnChange shouldMap consistency

motiazu:
In continuation to the discussion in https://github.com/acdlite/recompose/issues/666.
Saving prevProps returns the HOC to it's original behavior before the lifecycle updates.
Added test case that fails with the current implementation and passes with the fix.


istarkov:
Thank you! :-)

PR #713, opened 2018-07-26
Include license in published packages

taion:
Also, update the license dates.

For legal purposes, it's often preferable to have the actual license file. See also e.g. https://github.com/lerna/lerna/pull/1465.


istarkov:
Thank you!

Issue #717, opened 2018-07-30
Missing utils/ folder in recompose 0.28.0


goto-bus-stop:
When using eg. `onlyUpdateForKeys`, utils/pick.js is imported, but it does not exist: https://unpkg.com/recompose@0.28.0/utils/pick.js

It looks like the source files in utils/ are being output into the package root, because there is a pick.js there: https://unpkg.com/recompose@0.28.0/pick.js

The reason I think is that the Babel CLI gets each source file as a command line argument and gets a single output directory. The CLI doesn't rebase paths in that situation, it just puts all files in that directory. Maybe rather than a list of files, it should just get the src/packages/recompose as input with a recurse flag, using the --ignore flag to exclude tests and such? I'm not sure if v7 still supports that, but it might work.

I don't have time to look more into it this week so I'm sorry for not PRing a patch!

farism:
This PR may be related https://github.com/acdlite/recompose/pull/716

goto-bus-stop:
Thanks for the patch @farism!

PR #719, opened 2018-07-31
Preserve folder structure during build

farism:
Fixes https://github.com/acdlite/recompose/issues/717

By default babel only copies `*.js` files when passing in a whole directory. Instead of globbing for `*.js` files, just pass the entire `sourceDir` as input and configure babel to also ignore `node_modules`. This should preserve the folder structure.

I went to remove `glob` as a dep but couldn't find it listed as a dependency anywhere.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -7,7 +7,6 @@ const chalk = require('chalk')
 const { flowRight: compose } = require('lodash')
 const readline = require('readline-sync')
 const semver = require('semver')
-const glob = require('glob')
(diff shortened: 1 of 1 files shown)


istarkov:
Thank you

barbalex:
is it possible to quickly build a new npm version so people don't stumble over this when updating?

barbalex:
Weird thing is: This page does not have v28 in the releases. But v28 is latest on npm and that is what you get installed - which then errors.

Issue #602, opened 2018-02-02
Using event.stopPropagation() in withStateHandlers does not work.


philipnilsson:
```
const Component = withStateHandlers(
  () => ({}),
  {
     onUpdate: () => e => {
       e.stopPropagation();
     }
  }
);
```

### Expected
This code will stop event-propagation. 

### Actual
It doesn't

### Cause
Handlers are run asynchronously inside a `setState` closure, and so `stopPropagation` is not called on time.

### Solution
Change [this line](https://github.com/acdlite/recompose/blob/master/src/packages/recompose/withStateHandlers.js#L25) to run handlers eagerly, and not inside the closure.

istarkov:
This is expected. Line change will not help as it will break atomic setState updates. Use intermediate `withHandlers` to `stopPropagation` if needed or stop it directly in event handler ie `onClick={e => {e.stopPropagation(); handleBlabla(e);}}`
`withStateHandlers` is a flowtyped replacement of `withState` and not intended to have same abilities as `withHandlers`.
I'll close this as it's not a bug or anything bad.

wuct:
@philipnilsson thanks for your advice. I think swapping the order of the arguments do help. Besides, there is no need to call `event.persist()` inside `withStateHandlers` because now we can access the event synchronously. I've created a [code sandbox](https://codesandbox.io/s/vnm50y57xy) for this idea.

@istarkov how do you think?

ElvenMonky:
How come withStateHandlers was introduced with incorrect call order for stateUpdaters in the first place!

PR #734, opened 2018-08-29
Added third parameter to withStateHandlers …

ElvenMonky:
Added third parameter to withStateHandlers to allow better control over event handling (persistance and propagation) and over result passed to setState (either object or function). Second parameter left for backwards compatibility and can be set to null/undefined.


ElvenMonky:
With respect to proposal to change order of calls in issue #602

istarkov:
Lets make a new hoc, withStateUpdaters, + help, + tests, +flow types, +flow test

ElvenMonky:
Ok, seems done now. Sorry for the delay.

One thing to note: pushing from Windows didn't work. Had to fix it too

PR #744, opened 2018-10-01
Remove fbjs dependency

### LINKED ISSUE #742: Package size concern over fbjs
A picture worth a thousand words 2.74M is a lot (I'm trying to reduce Material-UI footprint):

capture d ecran 2018-09-24 a 19 25 23

https://packagephobia.now.sh/result?p=recompose

---

Most of this bytes seem to be coming from https://packagephobia.now.sh/result?p=fbjs. https://github.com/acdlite/recompose/blob/7d5adf7f49b21e3693f7d631551d282e5c8ef820/src/packages/recompose/shallowEqual.js#L1

I understand why it was added in the first place, to match React internals, but they have recently removed the fbjs dependency: https://github.com/facebook/react/pull/13046.
What do you think of using a lighter alternative?

@gaearon By any chance, do you have an advice for the replacement of shallow equal? Thanks!

shysteph:
Remove shallowEquals import from fbjs and just copy the function  #here
Fix #742 using solution from @gaearon (copy/paste)

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "lib/packages/recompose/dist/Recompose.umd.js": {
-    "bundled": 46555,
-    "minified": 16485,
-    "gzipped": 4623
+    "bundled": 46425,
+    "minified": 16484,
+    "gzipped": 4625
   },
   "lib/packages/recompose/dist/Recompose.min.js": {
-    "bundled": 42993,
-    "minified": 15205,
-    "gzipped": 4193
+    "bundled": 42863,
+    "minified": 15204,
+    "gzipped": 4194
   },
   "lib/packages/recompose/dist/Recompose.esm.js": {
-    "bundled": 30894,
-    "minified": 14729,
-    "gzipped": 3417,
+    "bundled": 32428,
+    "minified": 15083,
+    "gzipped": 3550,
     "treeshaked": {
       "rollup": {
-        "code": 340,
-        "import_statements": 340
+        "code": 310,
+        "import_statements": 310
       },
       "webpack": {
-        "code": 1869
+        "code": 1838
       }
     }
   }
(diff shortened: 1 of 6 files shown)


shysteph:
I had to update yarn.lock as well.  I have node 10 installed and upath 1.0.4 is incompatible.  I committed them as well because they are involved in the size snapshot.

istarkov:
Thank you!

oliviertassinari:
Awesome!

istarkov:
@oliviertassinari having that pure was commited into react corebase https://github.com/facebook/react/pull/13748 it is the latest days of recompose in material-ui ;-)

oliviertassinari:
Oh could Andrew do that to it's own project! We are still using `wrapDisplayName`, and it will takes before this new pure component is broadly available. We have recently inlined `geyDisplayName` for an IE11 compatibility use case.

istarkov:
Time goes fast :-) because of flow issues I didnt use recompose in my own projects more than 6 monthes. (relay also played a role as I almost dont need any data conversions)

oliviertassinari:
@istarkov recompose is damn useful. Both for authoring HOC and for consuming them. I love the new to render props helper.

MrLoh:
Will there be a new release soon, I am currently getting an error `"export 'default' (imported as 'shallowEqual') was not found in 'fbjs/lib/shallowEqual'`

MrLoh:
If I manually patch this commit in my node_modules, everything works fine again.

oliviertassinari:
@MrLoh Who is importing fbjs?

MrLoh:
I have no idea why this error occurs, but it comes from inside recompose (`node_modules/recompose/dist/Recompose.esm.js 90:14-26` to be exact). What matters however is that this also fixes the issue, but hasn't been released yet, so it cannot be used.

stof:
@istarkov any ETA for a release including that change ?

PR #745, opened 2018-10-04
Enable flow strict

dhui:
Allows downstream dependencies to enable the flow nonstrict-import rule


dhui:
[Official flow strict docs](https://flow.org/en/docs/strict/)

Without this change, any package that depends on recompose won't be able to use strict imports.

I also made a similar change to immutable-js: https://github.com/facebook/immutable-js/pull/1580

wuct:
Thanks

PR #746, opened 2018-10-06
Update README.md to reflect performance changes in React 16

sin:
As @trueadm wrote [in this issue](https://github.com/reactjs/reactjs.org/issues/639#issuecomment-367858928) on React documentation, the future performance optimizations mentioned in readme file are already there. 🎉 

> Functional components in React 16 don't go through the same code path as class components, unlike in previous version where they were converted to classes and would have the same code path. Class components have additional checks required and overhead in creating the instances that simple functions don't have.


wuct:
Thanks!

PR #747, opened 2018-10-06
Add version to allow yarn add from GitHub

MrLoh:
I am trying to install the master from GitHub, but yarn complains about a missing version


MrLoh:
This doesn't seem to be enough to build from GitHub, something with the build scripts seems to be configured wrongly

PR #751, opened 2018-10-17
composeWithDisplayName - a new utility function

arty-name:
I propose this change because I am unhappy with the current way we use `setDisplayName(wrapDisplayName(...))` when we want to wrap the name of the base component. 

Here’s one way to do that with an extra function wrapper: 
```js
const myHOC1 = BaseComponent => compose(
  setDisplayName(wrapDisplayName(BaseComponent, 'myName')),
  renameProp('old', 'new'),
  flattenProp('data'),
)(BaseComponent)
```

One could try to extract this functionality to avoid the wrapper: 
```js
const displayName = (Component) => 
  setDisplayName(wrapDisplayName(Component, 'myName'))(Component),

const myHOC2 = compose(
  displayName,
  renameProp('old', 'new'),
  flattenProp('data'),
)
```

However with this approach the `Component` already has display name `renameProp(flattenProp(...))` and we will just add yet another wrapper around it.

I propose a new utility function `composeWithDisplayName` to make the code more concise: 
```js
const myHOC3 = composeWithDisplayName(
  'myName',
  renameProp('old', 'new'),
  flattenProp('data'),
)
```

Have a look at the diff of this PR: many HOCs of recompose itself can benefit from this feature.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "lib/packages/recompose/dist/Recompose.umd.js": {
-    "bundled": 46425,
-    "minified": 16484,
-    "gzipped": 4625
+    "bundled": 45240,
+    "minified": 16419,
+    "gzipped": 4684
   },
   "lib/packages/recompose/dist/Recompose.min.js": {
-    "bundled": 42863,
-    "minified": 15204,
-    "gzipped": 4194
+    "bundled": 43923,
+    "minified": 15751,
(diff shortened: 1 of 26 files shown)


arty-name:
Here’s the [proposed documentation](https://github.com/acdlite/recompose/blob/5bd35e7e6bee4a48429fe01d5d05081e6fdb0fbc/docs/API.md#composewithdisplayname)

PR #752, opened 2018-10-18
Update API.md

salmanm:
Adjusted sample in docs to clarify what `composedHoc` is


istarkov:
current doc is right, compose of hocs is a function of BaseComponent

salmanm:
Ok makes sense. Thanks

PR #763, opened 2018-11-27
Fix hoistStatics hoisting statics of react ^16.3 features

eps1lon:
`hoist-non-react-statics@2.x` does not support `forwardRef` components, `contextType` static nor `getDerivedStateFromError` static. The major version bump was due to the added requirement of react >= 0.14 which was dropped in `hoist-non-react-statics@3.2.0`. Therefore this change should not be breaking.


noratarano:
@eps1lon @acdlite @wuct Any chance we can merge this? :) Would love to start using `contextType` along with `recompose`.

noratarano:
@eps1lon Would be great to update to version hoist-non-react-statics 3.3.0 now. ^_^

eps1lon:
I can do that but I have no permissions in this repository. I originally thought the deprecation would at least include bug fixes for a short amount of time but it seems like this isn't the case either. 

We used the `resolutions` field to force a higher version of `hoist-non-react-statics` but you would have to check if this is supported outside of `yarn`.

I'll update the PR again to use the highest version.

istarkov:
For now no maintener uses recompose in any env. In my case I havent used it more than 2 years. In such case its almost impossible to find any good to spend my free time on this project. As even this simple case is not just merging, it seriously increases library size, it has no tests - what we achieve adding this. React, flowtype and many others dev dependencies has changed a lot since last fixes, so Im sorry but it wont be merged.

eps1lon:
> so Im sorry but it wont be merged.

will it not be merged because of size increase or because of missing tests? Because I can add a test that is only testing a dependency if this is the only blocker. As to size increase: I would argue that a correct program is more important than a small program.

eps1lon:
Just checked: The size diff is because `.size-snapshot` is outdated on master. If it would be up-to-date then this PR wouldn't affect it.

noratarano:
@istarkov @eps1lon I'm not sure this PR demands much testing, if any. There are also no existing tests in the library checking that React statics are not copied, since I believe this is taken for granted since it's the purpose of the library: https://github.com/acdlite/recompose/blob/master/src/packages/recompose/hoistStatics.js

noratarano:
OK But anyway, just read this: https://github.com/acdlite/recompose#a-note-from-the-author-acdlite-oct-25-2018

inga-lovinde:
> OK But anyway, just read this: https://github.com/acdlite/recompose#a-note-from-the-author-acdlite-oct-25-2018

It says: "I will be discontinuing active maintenance of this package (excluding perhaps bugfixes or patches **for compatibility with future React releases**)"
This PR does exactly that.

PR #766, opened 2018-12-09
Fix: Use ref forwarding

mik-laj:
Hello, 

Currently, you can not get HTML credentials that go through this function. This creates a hard to notice bugs.

Thanks in advance.


wuct:
Hi @mik-laj, I am not sure what do you mean. Can you provide a test case which is failed with the current code?

PR #774, opened 2019-03-14
added $Exact to spread withStateHandlers properly

vaukalak:
To allow spread working properly, all it items should be $Exact types. Maybe previously it wasn't working this way, but in current version of flow, it's required.


wuct:
Can you provide a code example which shows the problem without `$Exact`? Thanks!

istarkov:
Also HOCs are broken with latests flow versions, better to wait until flow fixes that or not.

Issue #609, opened 2018-02-12
Any plans for the new context API?


FezVrasta:
Is there already some work done to support the upcoming context API?

It makes use of render props so I don't know how well will it fit with recompose, is this even going to be take in consideration?

istarkov:
lets discuss :-)

FezVrasta:
I'm far from an expert of HOCs, I think it's feasible, but I guess it will add 3 wrappers because of 1) the HOC, 2) the Context component, 3) the render prop

I'm not sure if this is going to be a problem.

wuct:
In my mind, it would be something like:

```
import { createContext } from 'recompose'

const themeContext = createContext({ color: 'green' })
```

`withContext` is a HOC wrapper for the context `Provider`, which accepts a function mapping props to context:

```
themeContext.withContext(props => ({ color: props.color }))
```

`getContext` is a HOC wrapper for the context `Consumer` (or we can rename it to `consumeContext`):

```
themeContext.getContext
```

istarkov:
Awesome @wuct . From my side I would like to add to getContext above an optional map function, so we could rename context variables.
Also I think that recompose createContext can also return default React consumer and provider.

wuct:
Looks good to me :)

SunHuawei:
Hi all, you may be interested in this. I have just published a npm package -- https://github.com/SunHuawei/with-context. I think it is the easiest way to use the new context API. 
Please try it and give me some feedback, really appreciate your attention.

wuct:
@SunHuawei the string key [here](https://github.com/SunHuawei/with-context/blob/master/index.js#L32) might be a pain in type systems.

liz-mars:
@SunHuawei  looks solid upon cursory glance. I wish that this was just how context worked in recompose

SunHuawei:
@wuct , thanks for your kind reminder. I'm not not sure what is the type systems you mentioned, the type system like flow/typescript, or PropTypes? However, any suggestion for that? I must have something to learn, honestly, could you show me the path?

wuct:
Using string keys means you can't determine what props will be pass to the base component until runtime, so you can't type the HOC with Flow/TypeScript. To avoid this problem, you can use a function instead. You can compare how we type `withStateHandlers` and `withState` in [flow](https://github.com/acdlite/recompose/blob/master/types/flow-typed/recompose_v0.24.x/flow_v0.55.x-/recompose_v0.24.x.js#L111) to get an idea.

[...]

PR #775, opened 2019-03-18
Change `withContext` and `getContext` to use new context API

illuminist:
related https://github.com/acdlite/recompose/issues/609

## Abstract

The new context API has an ability to bypass `PureComponent` and `shouldComponentUpdate` check that prevents component updating from context change, opposes to the legacy API that totally ignores context change if parent component haven't updated.

As far as I'm aware, this isn't breaking change, except some of those apps running by relying on the old API quirk. But overall this should improve the way the app behaves on context change.

The new context API rely on React 16.3 which is already a dependency of recompose package.

Here's a demo on how the new API behave compare to the legacy one https://codesandbox.io/s/1o5y0r419j

## The problem I found during build:

- I couldn't test locally because of jest kept throwing an error:
  ```
  SecurityError: localStorage is not available for opaque origins
  ```
  on every test since I first clone this repo. And I don't know how to fix this error.
- The test on 2 API might have been broken because the new context API require you to include its Provider/Consumer components which change the render tree.
- `yarn test` in `prepush` comment also has an error
  ```
  ERROR: [Errno 2] No such file or directory: 'test'
  ```

### DIFF
--- a/src/packages/recompose/contextManager.js
+++ b/src/packages/recompose/contextManager.js
@@ -1,0 +1,13 @@
+import { createContext } from 'react'
+
+const contextCache = {}
+
+export const getContextPair = contextName => {
+  if (contextCache[contextName]) {
+    return contextCache[contextName]
+  }
+  const contextPair = createContext()
+  contextPair.contextName = contextName
+  contextCache[contextName] = contextPair
+  return contextPair
+}
--- a/src/packages/recompose/getContext.js
+++ b/src/packages/recompose/getContext.js
@@ -1,16 +1,30 @@
-import { createFactory } from 'react'
+import React from 'react'
+import compose from './compose'
+import { getContextPair } from './contextManager'
 import setDisplayName from './setDisplayName'
 import wrapDisplayName from './wrapDisplayName'
+import mapValues from './utils/mapValues'
 
-const getContext = contextTypes => BaseComponent => {
-  const factory = createFactory(BaseComponent)
-  const GetContext = (ownerProps, context) =>
-    factory({
-      ...ownerProps,
-      ...context,
-    })
+const consumerHoc = (Consumer, contextName) => BaseComponent => props => (
+  <Consumer>
+    {value => <BaseComponent {...{ ...props, [contextName]: value }} />}
+  </Consumer>
+)
 
-  GetContext.contextTypes = contextTypes
+const getContext = childContextTypes => BaseComponent => {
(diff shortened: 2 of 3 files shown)


illuminist:
Turned out my OS has a very old system-wide yarn installation and the npm script tried to use it instead. So I fixed the problem by removing the old installation and installing the latest version via nvm I'm using.

Then came the next problem; `enzyme-adapter-react-16@1.1.0` doesn't support new React components, including new context API components. I had to change dependency version to `enzyme-adapter-react-16@^1.6.0` which is the current latest version. Then the tests for `getContext` and `withContext` runs fine now without any modification to the committed code.

Except for one test; `toClass`'s `toClass passes context and props correctly` still relies on old context API. I've never used `toClass` before, so I have no idea how it should behave or how it work.

So my questions are:
1. How should I refactor the `toClass` test to use new context API? Is it still needed?
2. I also had tried to refactor `withContext` to use jsx instead of `React.createElement`. But I'm not sure which one should be used; jsx or `React.createElement?

illuminist:
What I have changed are:

1. Refactor both component to produce a clearer render tree for multiple context Consumer/Provider component. Before it has
```
<WithContext ...props>
  <Provider value>
    <Component ...props>  <-- Unnecessary intermediate component
      <Provider value>
        <BaseComponent ...props>
```
After fixed:
```
<WithContext ...props>
  <Provider value>
    <Provider value>
      <BaseComponent ...props>
```
2. Use class component for `withContext` and add legacy API to it. This had be done to pass the `toClass` test that check context with legacy context API. This also add backward compatibility in case of some user uses recompose context component MIXED with classic way to manage context, which is also a bad practice.

3. Bump up `enzyme-adapter-react-16@^1.6.0`. New version is required to test new react component; context Provider/Consumer.

istarkov:
Great work, but see main page notice https://github.com/acdlite/recompose#a-note-from-the-author-acdlite-oct-25-2018
there is no active maintenance of this package any longer, use hooks, use native React context, its much more easier vs recompose.

illuminist:
I understand that. I really love to use hooks, but the only reason I haven't used full hooks is my app mainly uses redux and a hook for redux isn't ready yet as they're still discussing in https://github.com/reduxjs/react-redux/issues/1177. And I don't want to use https://github.com/facebookincubator/redux-react-hook just to be migrated again after official hook is ready.

PR #826, opened 2021-02-03
Modernize recompose

### LINKED ISSUE #806: Remove usage of `createFactory`
In react@16.13.1 this library triggers a warning:

> Warning: React.createFactory() is deprecated and will be removed in a future major release. Consider using JSX or use React.createElement() directly instead.

It looks like this repository [uses this API in quite a few places](https://github.com/acdlite/recompose/search?q=createFactory&unscoped_q=createFactory), so perhaps this should be refactored?

brodybits:
Starting with switch to internal __createFactory function to resolve #806

Note that I would like to find a way to test the `dist` artifacts before updating the build-related dependencies.

### DIFF
--- a/src/packages/recompose/createFactory.js
+++ b/src/packages/recompose/createFactory.js
@@ -1,0 +1,9 @@
+import { createElement } from 'react'
+
+// NOTE: This is intended to be an internal utility for now, may be
+// unit-tested, documented, and exported by the API in the future.
+export default function __createFactory(c) {
+  // as described in:
+  // - https://reactjs.org/blog/2020/02/26/react-v16.13.0.html
+  return createElement.bind(null, c)
+}
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,18 +1,18 @@
 {
   "lib/packages/recompose/dist/Recompose.umd.js": {
-    "bundled": 46425,
-    "minified": 16484,
-    "gzipped": 4625
+    "bundled": 46573,
+    "minified": 16272,
+    "gzipped": 4626
   },
   "lib/packages/recompose/dist/Recompose.min.js": {
-    "bundled": 42863,
-    "minified": 15204,
-    "gzipped": 4194
+    "bundled": 43011,
+    "minified": 14992,
+    "gzipped": 4185
   },
   "lib/packages/recompose/dist/Recompose.esm.js": {
-    "bundled": 32428,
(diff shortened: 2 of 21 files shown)


src/packages/recompose/componentFromStream.js
```diff
@@ -29,7 +29,7 @@ export const componentFromStreamWithConfig = config => propsToVdom =>
     // Stream of vdom
     vdom$ = config.toESObservable(propsToVdom(this.props$))
 
-    componentWillMount() {
+    componentDidMount() {
```
brodybits, line 32: I am not sure if this change, which is probably the safest way to support multiple React versions and forks, could lead to potential loss of optimization or any other potential issues.

src/packages/recompose/componentFromStream.js
```diff
@@ -39,6 +39,9 @@ export const componentFromStreamWithConfig = config => propsToVdom =>
       this.propsEmitter.emit(this.props)
     }
 
+    // TODO: componentWillReceiveProps method is not supported by React 17;
+    // UNSAFE_componentWillReceiveProps may not work with older React versions.
+    // Another solution is wanted.
     componentWillReceiveProps(nextProps) {
```
brodybits, line 45: TODO item

src/packages/recompose/createFactory.js
```diff
@@ -0,0 +1,9 @@
+import { createElement } from 'react'
+
+// NOTE: This is intended to be an internal utility for now, may be
+// unit-tested, documented, and exported by the API in the future.
+export default function __createFactory(c) {
```
brodybits, line 5: I think it would be nice to export and document this in the API.

src/packages/recompose/utils/mapValues.js
```diff
@@ -1,12 +1,14 @@
 const mapValues = (obj, func) => {
   const result = {}
-  /* eslint-disable no-restricted-syntax */
+  // TODO: no-unused-vars seems to trigger a issue in recent eslint versions
+  /* eslint-disable no-restricted-syntax, no-unused-vars */
   for (const key in obj) {
     if (obj.hasOwnProperty(key)) {
       result[key] = func(obj[key], key)
```
brodybits, line 7: I would like to find a nicer, cleaner solution for this without potentially breaking ES5 consumers.

brodybits (commented):
Making note of some TODO items

